PR #6, opened 2016-10-31
Implement getItemsForUser

TimMoore:
- Add a Cassandra read side processor for `PItemEvent`s that allows for querying by creator and status.
- Create a profile UI in the web gateway for displaying the current user's item in each status.
- Implement the service call.

### DIFF
--- a/item-api/src/main/java/com/example/auction/item/api/ItemService.java
+++ b/item-api/src/main/java/com/example/auction/item/api/ItemService.java
@@ -21,67 +21,71 @@ import static com.lightbend.lagom.javadsl.api.Service.*;
 
 /**
  * The item service.
- *
  * Manages the lifecycle of items, as well properties on them.
  */
 public interface ItemService extends Service {
 
-  /**
-   * Create an item.
-   *
-   * @return The created item with its ID populated.
-   */
-  ServiceCall<Item, Item> createItem();
+    /**
+     * Create an item.
+     *
+     * @return The created item with its ID populated.
+     */
+    ServiceCall<Item, Item> createItem();
 
-  /**
-   * Update an item.
-   *
-   * @param id The ID of the item to update.
-   * @return Done.
-   */
-  ServiceCall<Item, Done> updateItem(UUID id);
+    /**
+     * Update an item.
+     *
+     * @param id The ID of the item to update.
+     * @return Done.
+     */
+    ServiceCall<Item, Done> updateItem(UUID id);
 
-  /**
-   * Start an auction for an item.
-   *
-   * @param id The id of the item to start the auction for.
-   * @return Done if the auction was started.
-   */
-  ServiceCall<NotUsed, Done> startAuction(UUID id);
+    /**
+     * Start an auction for an item.
+     *
+     * @param id The id of the item to start the auction for.
+     * @return Done if the auction was started.
+     */
+    ServiceCall<NotUsed, Done> startAuction(UUID id);
 
-  /**
-   * Get an item with the given ID.
-   *
-   * @param id The ID of the item to get.
-   * @return The item.
-   */
-  ServiceCall<NotUsed, Item> getItem(UUID id);
+    /**
+     * Get an item with the given ID.
+     *
+     * @param id The ID of the item to get.
+     * @return The item.
+     */
+    ServiceCall<NotUsed, Item> getItem(UUID id);
 
-  /**
-   * Get a list of items for the given user.
-   *
-   * @param id The ID of the user.
-   * @param pageNo The page number.
-   * @param pageSize The number of items to return per page.
-   * @return The sequence of items.
-   */
-  ServiceCall<NotUsed, PSequence<Item>> getItemsForUser(UUID id, Optional<Integer> pageNo, Optional<Integer> pageSize);
+    /**
+     * Get a list of items for the given user.
+     *
+     * @param id       The ID of the user.
+     * @param status   The status of items to return.
+     * @param pageNo   The page number, starting from zero.
+     * @param pageSize The number of items to return per page.
+     * @return The sequence of items.
+     */
+    ServiceCall<NotUsed, PaginatedSequence<ItemSummary>> getItemsForUser(
+            UUID id, ItemStatus status, Optional<Integer> pageNo, Optional<Integer> pageSize);
 
-  /**
-   * The item events stream.
-   */
-  Topic<ItemEvent> itemEvents();
+    /**
+     * The item events stream.
+     */
+    Topic<ItemEvent> itemEvents();
 
-  @Override
-  default Descriptor descriptor() {
-    return named("item").withCalls(
-            pathCall("/api/item", this::createItem),
-            restCall(Method.POST, "/api/item/:id/start", this::startAuction),
-            pathCall("/api/item/:id", this::getItem),
-            pathCall("/api/item?userId&pageNo&pageSize", this::getItemsForUser)
-    ).publishing(
-            topic("item-ItemEvent", this::itemEvents)
-    ).withPathParamSerializer(UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString))
-            .withHeaderFilter(SecurityHeaderFilter.INSTANCE);
-  }
+    @Override
+    default Descriptor descriptor() {
+        return named("item").withCalls(
+                pathCall("/api/item", this::createItem),
+                restCall(Method.POST, "/api/item/:id/start", this::startAuction),
+                pathCall("/api/item/:id", this::getItem),
+                pathCall("/api/item?userId&status&pageNo&pageSize", this::getItemsForUser)
+        ).publishing(
+                topic("item-ItemEvent", this::itemEvents)
+        ).withPathParamSerializer(
+                UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString)
+        ).withPathParamSerializer(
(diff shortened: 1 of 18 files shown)


TimMoore:
This project isn't really set up with tests, so I'll add some in a separate PR

item-api/src/main/java/com/example/auction/item/api/ItemService.java
```diff
@@ -21,67 +21,71 @@
...
+                pathCall("/api/item", this::createItem),
+                restCall(Method.POST, "/api/item/:id/start", this::startAuction),
+                pathCall("/api/item/:id", this::getItem),
+                pathCall("/api/item?userId&status&pageNo&pageSize", this::getItemsForUser)
+        ).publishing(
+                topic("item-ItemEvent", this::itemEvents)
+        ).withPathParamSerializer(
+                UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString)
+        ).withPathParamSerializer(
+                ItemStatus.class, PathParamSerializers.required("ItemStatus", ItemStatus::valueOf, ItemStatus::toString)
```
TimMoore: It would be nice if enums could be serialized/deserialized without extra configuration. Something to add to the core maybe?

item-impl/src/main/java/com/example/auction/item/impl/CompletionStageUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+final class CompletionStageUtils {
```
TimMoore: These are a bunch of functions that I found useful when writing the read side processor. Maybe some of them should be considered for promotion to the core.

item-impl/src/main/java/com/example/auction/item/impl/CompletionStageUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        throw new Error("no instances");
+    }
+
+    static <R, X extends RuntimeException> Function<Optional<R>, R> throwIfEmpty(Supplier<X> exceptionThrower) {
+        return result -> result.orElseThrow(exceptionThrower);
+    }
+
+    static CompletionStage<List<BoundStatement>> completedStatements(BoundStatement... statements) {
+        return CassandraReadSide.completedStatements(Arrays.asList(statements));
+    }
```
TimMoore: This one in particular seems like a pretty natural convenience to add to the ones that are already there.

item-impl/src/main/java/com/example/auction/item/impl/CompletionStageUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+    static <T> Function<T, Done> accept(Consumer<T> f) {
+        return t -> {
+            f.accept(t);
+            return Done.getInstance();
+        };
+    }
+
+    static CompletionStage<Done> doAll(CompletionStage<?>... stages) {
```
TimMoore: Surprised that something like this one isn't in the JDK. There is a `CompletableFuture.allOf` that is similar, but it only works with `CompletableFuture` instances and not arbitrary `CompletionStage` for some reason.

Also, this of course uses `Done` instead of `Void` to conform with the Lagom expectations... why _do_ we use `Done` instead of `Void` anyway? It makes a few things more awkward.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -0,0 +1,248 @@
...
+
+@Singleton
+public class ItemRepository {
+
+    private final CassandraSession session;
+
+    @Inject
+    public ItemRepository(CassandraSession session, ReadSide readSide) {
+        this.session = session;
+        session.underlying().thenAccept(s -> registerCodec(s, new EnumNameCodec<>(ItemStatus.class)));
```
TimMoore: Is this a good place to register codecs for the Cassandra Java driver? It might be nice to offer a convenience for that.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -0,0 +1,248 @@
...
+        return session
+                .selectOne(
+                        "SELECT COUNT(*) FROM itemSummaryByCreatorAndStatus " +
+                                "WHERE creatorId = ? AND status = ? " +
+                                "ORDER BY status ASC, itemId DESC",
+                        // ORDER BY status is required due to https://issues.apache.org/jira/browse/CASSANDRA-10271
+                        creatorId,
+                        status
+                )
+                .thenApply(row -> (int) row.get().getLong("count"));
```
TimMoore: It returns `count` as a long, but only allows `LIMIT` as an int!

It's a bit of a cheat, but if there are ever more than `Integer.MAX_VALUE` results, pagination will be completely broken 😕

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -0,0 +1,248 @@
...
+                                    "WITH CLUSTERING ORDER BY (itemId DESC)"
+                    ).thenCompose(done ->
+                            session.executeCreateTable(
+                                    "CREATE MATERIALIZED VIEW IF NOT EXISTS itemSummaryByCreatorAndStatus AS " +
+                                            "SELECT * FROM itemSummaryByCreator " +
+                                            "WHERE status IS NOT NULL AND itemId IS NOT NULL " +
+                                            "PRIMARY KEY (creatorId, status, itemId) " +
+                                            "WITH CLUSTERING ORDER BY (status ASC, itemId DESC)"
+                            )
+                    )
```
TimMoore: A subtlety here is that the two table creation statements can execute potentially in parallel or in an arbitrary order because there's no dependency between them, but the materialized view does depend on the second table creation.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
```diff
@@ -66,7 +68,7 @@ public ItemServiceImpl(PersistentEntityRegistry registry, BiddingService bidding
             if (!userId.equals(item.getCreator())) {
                 throw new Forbidden("User " + userId + " can't created an item on behalf of " + item.getCreator());
             }
-            UUID itemId = UUID.randomUUID();
+            UUID itemId = UUIDs.timeBased();
```
TimMoore: This gives `itemId` a well-defined ordering.

item-api/src/main/java/com/example/auction/item/api/ItemService.java
jroper: Yeah - though currently lookup of serializers is done via class match, not via class heirarchy.  I think though it's all done ahead of time, so there's no harm in doing a dumb iteration through all classes and doing a `isAssignableFrom` or whatever. We should just do the exact class match first.

item-impl/src/main/java/com/example/auction/item/impl/CompletionStageUtils.java
jroper: Really?  There's no varargs method?  Ok.

item-impl/src/main/java/com/example/auction/item/impl/CompletionStageUtils.java
jroper: There's a few reasons.  Firstly, `Void` is not instantiatable, so you must pass `null`.  This causes problems when interacting with other APIs that don't support `null`, such as `Optional.of`, and reactive streams also doesn't allow `null` elements.  Secondly, `Void` doesn't really communicate the intent.  Akka actually introduced two different types `NotUsed` and `Done`.  `NotUsed` is more semantically equivalent to `Void`, it's used where something has to have a parameter type, but that parameter is not used, and so whatever is passed there has no meaning.  In contrast, if something gives you a `Done`, you know whatever it was doing is now done.  A good example of where this distinction matters is if you have `ServiceCall<Source<SomeMessage, ?>, NotUsed>` vs `ServiceCall<Source<SomeMessage, ?>, Done>`. A service call of that shape currently gets implemented using websockets, with the return message being a single message.  If the type is `NotUsed`, Lagom will not wait for a return message, it will immediately return `NotUsed`.  But if it's `Done`, then it treats it as a regular message, and doesn't emit anything until it gets a `Done` message back (which by default is implemented by passing an empty message).

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
jroper: I think so.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
jroper: Let's hope we have some user quota API that prevents a single user from creating 2 billion auctions...

item-api/src/main/java/com/example/auction/item/api/ItemSummary.java
```diff
@@ -0,0 +1,44 @@
+package com.example.auction.item.api;
+
+import java.util.UUID;
+
+/**
+ * The most important fields of an item.
+ */
+public final class ItemSummary {
```
TimMoore: Raised as #9

PR #11, opened 2016-11-23
Integration tests on item service

ignasi35:
Increases coverage in ItemService. There's unit tests at PersistentEntity level, ReadSide level and integration test at Service level.

Introduces OptionalSerializer, Await and ReadSideTestDriver (all WIP)


ignasi35:
Travis fails on ReadSideTestDriver wiring. Increased timeout...

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
```diff
@@ -43,25 +44,32 @@ public ItemServiceImpl(PersistentEntityRegistry registry, BiddingService bidding
...
+        biddingService.bidEvents().subscribe().atLeastOnce(Flow.<BidEvent>create().mapAsync(1, this::handleBidEvent));
     }
 
-    
+  /**
+   *  TODO: doc this.
+   * @param event
+   * @return
+   */
+    private CompletionStage<Done> handleBidEvent(BidEvent event) {
```
TimMoore: Good idea extracting this, and I might go one step further and extract each event handler to a separate method as well.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
```diff
@@ -51,7 +52,7 @@ private Behavior created(PItemState state) {
 
         builder.setCommandHandler(StartAuction.class, (start, ctx) -> {
             if (start.getUserId().equals(state().getItem().get().getCreator())) {
-                return ctx.thenPersist(new AuctionStarted(entityUuid(), Instant.now()), evt -> ctx.reply(Done.getInstance()));
+                return ctx.thenPersist(new AuctionStarted(entityUuid(), Instant.now()), done(ctx));
```
TimMoore: Think it might be good to implement `ctx.done()` in the framework as a convenience.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
```diff
@@ -103,14 +108,26 @@ private Behavior cancelled(PItemState state) {
...
+        builder.setReadOnlyCommandHandler(UpdatePrice.class, this::alreadyDone);
+        builder.setReadOnlyCommandHandler(FinishAuction.class, this::alreadyDone);
 
         return builder.build();
     }
 
+    /**
+     * Convenience method to handle when a command has already been processed (idempotent processing).
+     */
+    private void alreadyDone(Object command, ReadOnlyCommandContext<Done> ctx) {
```
TimMoore: It might be better to call this `ignore`... what do you think?

Similar to above, it would be good to pull that up into `PersistentEntity` as a convenience.

item-impl/src/main/java/com/example/auction/item/impl/PItemState.java
```diff
@@ -54,4 +54,19 @@ public PItemStatus getStatus() {
         return item.map(PItem::getStatus).orElse(PItemStatus.NOT_CREATED);
     }
```
TimMoore: With the addition of `OptionalSerializer`, this whole class may now be unnecessary.

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
```diff
@@ -0,0 +1,85 @@
...
+    AggregateEventTagger<Event> tag = e.aggregateTag();
+
+    List<Pair<ReadSideProcessor.ReadSideHandler<?>, Offset>> list = processors.get(tag.eventType());
+
+    if (list == null) {
+      throw new RuntimeException("No processor registered for Event " + tag.eventType().getCanonicalName());
+    }
+
+    List<CompletionStage<Done>> stages = list.stream().map(pHandlerOffset -> {
+          Flow<Pair<Event, Offset>, Done, ?> flow = ((ReadSideProcessor.ReadSideHandler<Event>) pHandlerOffset.first()).handle();
```
TimMoore: Why does it need a cast here?

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
```diff
@@ -0,0 +1,85 @@
...
+          return Source.single(Pair.create(e, offset)).via(flow).runWith(Sink.ignore(), materializer);
+        }
+    ).collect(Collectors.toList());
+    return CompletionStageUtils.doAll(stages);
```
TimMoore: `doAll` does not sequence the stages, it (potentially) runs them all concurrently. Is that what we want here? I think we might need to run them in order.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
ignasi35: I'm not sure I like `ignore` (in hindsight I don't like `alreadyDone`either :) ). While there's no processing happening, the method is side-effecting a `Done.getInstance()` back through the `ctx`. 

Overload `done` ? `doNothing` ? `identity` ?

I like the idea of promoting the method into `PersistentEntity` better than adding a `ctx.done` (see your previous comment.

item-impl/src/main/java/com/example/auction/item/impl/PItemState.java
ignasi35: Good catch. Will rewrite.

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
ignasi35: Will review. Most probably IntelliJ messing up type inference.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
TimMoore: It appears that `ctx.done` already exists! http://www.lagomframework.com/documentation/1.2.x/java/api/com/lightbend/lagom/javadsl/persistence/PersistentEntity.CommandContext.html#done--

However it doesn't do the same thing.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
TimMoore: I suggested "ignore" because that's the word you used in the comments.

To be honest, I'm now kind of confused about when it is necessary to send a reply. Is it always necessary? Then why doesn't the API enforce this, and what is the use of the existing `ctx.done` method?

ignasi35: Well, if the command defines a ReplyType and the service is waiting on the response ( `ask` ) then replying is a must, right? From the client PoV, it's irrelevant if the message was processed or ignored due to repetition because we're covering a command retrial. Also, this `ignore` method could only be used when the ReplyType is `ReplyType<Done>` (but it's a very common case).

Back to `ctx.done`, I agree with you that we could simply not sent events and keep going.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
ignasi35: We did that one together, remember? LOL

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
ignasi35: Also, in https://github.com/lagom/online-auction-java/blob/master/bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionEntity.java we have another `alreadyDone` (I knew I had copied from somewhere).

item-impl/src/main/java/com/example/auction/item/impl/PItemState.java
ignasi35: I see you point but `PItemState` in particular wold be hard to get rid of. `PItemState` holds a `PItem` and a `PItemStatus` and also provides few methods to make state transitions during the auction easier. But I'll keep an eye on anaemic States that could be replaced with an `Optional<T>`.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
TimMoore: Haha clearly my memory is not very good!

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
ignasi35: Not IntelliJ. SuperInterface requires `register` to define a method-scope Generic type. Then, on `feed()`we're using a different method-scope Generic type. Both methods have access to a map that's typed wildcard. I don't see a quick/easy way out. :-(

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
TimMoore: I see... I still can't help but think that there could be a simpler API. I'll keep thinking about it.

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
ignasi35: This is running the event fed into the test-driver into all the processors registered to that event type. I think we're good if all processors are run in parallel.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
ignasi35: Agreed. I'll rename all to `alreadyDone` for consistency and we can review when moving it to lagom core.

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
TimMoore: ah yes I see

item-impl/src/test/java/com/example/auction/item/impl/testkit/ReadSideTestDriver.java
TimMoore: right, got it

ignasi35:
Build keeps failing even if using a Timeout of 10 seconds in `ReadSideTestDriver` when registering processors. Will bump timeout once again.

ignasi35:
Let's merge.

TimMoore:
Hmm I seem to have lost (or never had?) my merge permission. @jroper?

PR #14, opened 2016-11-29
Edit item

ignasi35:
Adds feature to edit items. Only the creator may edit an item and an item may be edited while the auction is not open (during auction the description can be edited).


build.sbt
```diff
@@ -156,4 +156,4 @@ lazy val eclipseSettings = Seq(
   unmanagedSourceDirectories in Test := Seq((javaSource in Test).value)
 )
 
-lagomCassandraCleanOnStart in ThisBuild := false
+lagomCassandraCleanOnStart in ThisBuild := true
```
TimMoore: Why make this change? Did you mean to commit it?

item-api/src/main/java/com/example/auction/item/api/ItemService.java
```diff
@@ -73,19 +72,25 @@
...
 
+
+
     @Override
     default Descriptor descriptor() {
         return named("item").withCalls(
                 pathCall("/api/item", this::createItem),
                 restCall(Method.POST, "/api/item/:id/start", this::startAuction),
                 pathCall("/api/item/:id", this::getItem),
+                pathCall("/api/item/:id/update", this::updateItem),
```
TimMoore: I think it would be more idiomatic to handle this as a `POST` or `PUT` to `/api/item/:id`.

item-api/src/main/java/com/example/auction/item/api/ItemService.java
```diff
@@ -73,19 +72,25 @@
...
                 pathCall("/api/item?userId&status&pageNo&pageSize", this::getItemsForUser)
         ).publishing(
                 topic("item-ItemEvent", this::itemEvents)
         ).withPathParamSerializer(
                 UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString)
         ).withPathParamSerializer(
                 ItemStatus.class, PathParamSerializers.required("ItemStatus", ItemStatus::valueOf, ItemStatus::toString)
+        ).withPathParamSerializer(
+                UpdateItemResultCodes.class, PathParamSerializers.required("UpdateItemResultCode", UpdateItemResultCodes::valueOf, UpdateItemResultCodes::toString)
```
TimMoore: Is this necessary? It doesn't seem to be used.

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
```diff
@@ -43,6 +41,92 @@ public int hashCode() {
         }
     }
 
+
+    final class UpdateItem implements PItemCommand, PersistentEntity.ReplyType<PUpdateItemResult> {
+        private final UUID id;
+        private final UUID creator;
```
TimMoore: I don't think we want to allow the creator to be updated

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
```diff
@@ -43,6 +41,92 @@ public int hashCode() {
...
+        final String title;
```
TimMoore: Is there a specific reason why `title` isn't private when everything else is?

item-api/src/main/java/com/example/auction/item/api/UpdateItemResult.java
```diff
@@ -0,0 +1,31 @@
+package com.example.auction.item.api;
+
+
+public final class UpdateItemResult {
```
TimMoore: It might be good to send the updated item state back in the result as well.

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
```diff
@@ -43,6 +41,92 @@ public int hashCode() {
...
+        private final String description;
+        private final String currencyId;
+        private final int increment;
+        private final int reservePrice;
+        private final Duration auctionDuration;
+
+        @JsonCreator
+        public UpdateItem(UUID id, UUID creator, String title, String description, String currencyId, int increment, int reservePrice, Duration auctionDuration) {
```
TimMoore: We are really duplicating the `Item`/`PItem` details into a lot of different places, which is going to make it very painful to add fields. I wonder whether we should have the command simply wrap a PItem derived from the item provided in the request. That's how `CreateItem` works.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
```diff
@@ -50,6 +53,14 @@ private Behavior created(PItemState state) {
 
         builder.setReadOnlyCommandHandler(GetItem.class, this::getItem);
 
+        builder.setCommandHandler(UpdateItem.class, (updateItem, ctx) -> {
+                    ItemUpdated event = ItemUpdated.from(updateItem);
```
TimMoore: I wonder if there's any additional validation we should do on the item passed in.

For example:

- We should reject the command if the creator has changed
- We should not allow the status to change via this command (or, if we do, we need to be sure to trigger the appropriate behavior)
- Clients should not be able to change start/end dates or winner via this method, either
- If _none_ of the fields have changed, it should not emit an event

Some of these are handled currently by simply not copying the fields across, but this has the drawback that the client does not know which fields that it sent in the request actually changed, or that it sent a (partially) invalid request. This could cause a client to get out of date with the server.

This is another good reason to include all of the details from the original item on the request (so we have all of the fields available to validate) and to return the current state of the item in the response (so that clients can resynchronize their local state if necessary).

(Arguably, a lot of this should apply to create as well.)

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
```diff
@@ -90,15 +113,19 @@ private Behavior completed(PItemState state) {
 
         builder.setReadOnlyCommandHandler(GetItem.class, this::getItem);
 
+        // a completed auctio's item can't be edited.
+        builder.setReadOnlyCommandHandler(UpdateItem.class, (updateItem, ctx) ->
+                ctx.reply(new PUpdateItemResult(UpdateItemResultCodes.CANT_UPDATE_AUCTION_IS_CLOSED))
```
TimMoore: Why not use `invalidCommand` here?

item-impl/src/main/java/com/example/auction/item/impl/PItemEvent.java
```diff
@@ -57,6 +58,104 @@ public String toString() {
         }
     }
 
+    final class ItemUpdated implements PItemEvent {
```
TimMoore: Again, I think it would be simpler to wrap the `PItem`.

web-gateway/app/controllers/Durations.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class Durations {
+
+    public static Pair<ChronoUnit, Long> fromJDuration(Duration duration) {
+        if (duration.getSeconds() % ChronoUnit.WEEKS.getDuration().getSeconds() == 0) {
+            return Pair.create(ChronoUnit.WEEKS, duration.toDays() / 7);
+        } else if (duration.getSeconds() % ChronoUnit.DAYS.getDuration().getSeconds() == 0) {
+            return Pair.create(ChronoUnit.DAYS, duration.toDays());
+        } else if (duration.getSeconds() % ChronoUnit.HALF_DAYS.getDuration().getSeconds() == 0) {
+            return Pair.create(ChronoUnit.HALF_DAYS, duration.toHours() / 12);
```
TimMoore: Half days are a bit of an odd unit (and are not handled in `Nav` below. Should we just treat these as hours?

web-gateway/app/views/editItem.scala.html
```diff
@@ -1,14 +1,30 @@
 @import helper._
```
TimMoore: Is there some way to reduce the duplication between create & edit forms?

web-gateway/app/views/editItem.scala.html
```diff
@@ -1,14 +1,30 @@
...
-    <h2>@message("createItem")</h2>
 
-    @foundationForm(itemForm, routes.ItemController.createItem()) {
+@main(message("editItem")) {
+
+    <h2>@message("editItem")</h2>
+
+    @foundationForm(itemForm, routes.ItemController.editItem(itemId.toString)) {
+
+        @if(errorMessage.isPresent) {
```
TimMoore: Definitely necessary to handle errors from the server, but I think it ought to also check the current state of the item when the form is rendered and disable fields that shouldn't be editable.

build.sbt
ignasi35: No. My bad.

item-api/src/main/java/com/example/auction/item/api/ItemService.java
ignasi35: I don't know why I comited that. 😭 I was in rage trying to fix a serialization issue and removed that.

ignasi35: it is a `POST` ;-)

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
ignasi35: We don't. I just needed those two fields on readside and added them on the event. Then... intertia kicked in and added them on the command too. Fixing. Thanks.

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
ignasi35: None. fixing.

item-impl/src/main/java/com/example/auction/item/impl/PItemCommand.java
ignasi35: `CreateItem` has the advantage that is only handled when there's no `PItemState`. Updating, OTH, needs a previous `PItemState` (with it's internal `PItem`) and, on top of that, can happen on `create` and on `auction` behaviours.

item-api/src/main/java/com/example/auction/item/api/UpdateItemResult.java
ignasi35: I thought of responding  with a `Location:` header but that'd be 2 round trips. I'll send the item back. 👍

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
ignasi35: I was (wrongly) asserting on the service that the principal of the update action was equal to the creatorId in the payload of the request. I will change it so that the principal is considered the _commander_ on the  `UpdateItem`. Then the PersistentEntity can assert that the `creator` matches the `commander`.
By not using `Pitem` as the payload of `UpdateItem` many of the checks you suggest are not necessary.

I did forget to block the event generation when no field changes.

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
ignasi35: I don't want an exception sent back. I could capture it on the service though...

Overall I'm not very happy with the error handling ATM.

web-gateway/app/views/editItem.scala.html
ignasi35: They are diverging and I think they will continue to diverge. Let's see what can we reuse in a few days.

web-gateway/app/views/editItem.scala.html
ignasi35: True. Should only allow editing Description when Auction is ongoing.

ignasi35:
I extracted from `PItem` those fields that may be edited from the form. Those 6 fields were showing up as a group over and over. This new Tuple6 is named `PItemDetails` and has a different lifecycle than `AuctionStatus`, winners, start and end. This extraction also helps preventing the update of fields not available to the end user.
When editing I used the concept of a `commander` as a principal authoring a command. On the `PItemEntity` it is now possible to validate that the update command was requested (commanded) by the same person that created the entity. This value is captured from the request principal (it is not read from the request payload).

When updating, as @TimMoore pointed out, there's event triggered only when a field was changed.


PS: My twirl fu is very poor, and I think the `editItem` template could be improved. Any pointers will be welcome but ATM I'm comfortable with it. I'm not sure we can reuse much code from `editItem` to `createItem`.

item-api/src/main/java/com/example/auction/item/api/ItemService.java
jroper: Yes, but path param serializers are only needed for things that appear in the path (URL) of a request.  If the result code only appears in a message body (either a request message or a response message), a path param is not needed.  The item status path param serializer was needed for example because it's part of the query string when searching for items by user. But I don't see the update item result code ever being used in a URL.

item-api/src/main/java/com/example/auction/item/api/UpdateItemResultCodes.java
```diff
@@ -0,0 +1,11 @@
+package com.example.auction.item.api;
+
+/**
+ *
+ */
+public enum UpdateItemResultCodes {
+    NOT_FOUND,
+    SUCCESS,
+    CAN_ONLY_UPDATE_DESCRIPTION,
+    CANT_UPDATE_AUCTION_IS_CLOSED;
```
jroper: I don't think I'd use an enum here.  There's a few reasons:

1. There's only one success result. This is different to placing a bid, where there are multiple different success results, your bid may have been accepted (success) but was below the reserve, your bid may have been accepted (success) but was automatically outbid, and your bid may have been accepted (success) and is the current highest bid.
2. The UI is already going to render in such a way that it should only allow editing fields that are editable. So there's not a big need for describing exactly why it failed, the chance of a user encountering these errors is very low because the UI ensures this.
3. Most of these status can be expressed semantically in HTTP - NOT FOUND can be expressed by returning a 404 error, SUCCESS by returning 200, and the other two by returning some other 4xx error.  I think it's better to use semantic HTTP status codes than inventing a new protocol for returning the status.

By the way, you can easily return any status code with a message by throwing a TransportException, eg:

```java
throw new TransportException(TransportErrorCode.fromHttp(409), "Can only edit description once the auction has started")
```

item-api/src/main/java/com/example/auction/item/api/ItemService.java
TimMoore: But I mean that creating a separate "update" path is not really the way REST is typically expressed

item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java
TimMoore: Why don't you want an exception sent back? It seems like an exceptional situation.

web-gateway/app/views/editItem.scala.html
TimMoore: OK... I suspect there will still be some commonality remaining 😄

item-impl/src/main/java/com/example/auction/item/impl/PItem.java
```diff
@@ -12,32 +12,35 @@
...
+        this.creator = creator;
+        this.itemDetails = pItemDetails;
+        this.price = price;
+        this.status = status;
+        this.auctionStart = auctionStart;
+        this.auctionEnd = auctionEnd;
+        this.auctionWinner = auctionWinner;
+    }
+
+    @Deprecated
```
TimMoore: Why deprecate? It's an implementation class, not part of the API, so could just go ahead and remove it.

item-api/src/main/java/com/example/auction/item/api/PaginatedSequence.java
```diff
@@ -1,15 +1,17 @@
...
+@Value
 public class PaginatedSequence<T> {
-    private final PSequence<T> items;
-    private final int page;
-    private final int pageSize;
-    private final int count;
+    PSequence<T> items;
+    int page;
+    int pageSize;
+    int count;
```
ignasi35: Using Lombok's `@Value` makes it unnecessary to mark fields `private final`.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -182,15 +186,15 @@ public PItemEventProcessor(CassandraSession session, CassandraReadSide readSide)
         private CompletionStage<Done> prepareInsertItemSummaryByCreatorStatement() {
             return session.
                     prepare("INSERT INTO itemSummaryByCreator(" +
-                            "creatorId, " +
                             "itemId, " +
+                            "creatorId, " +
                             "title, " +
```
ignasi35: Throughout all the code, `itemId` precedes `creatorId`.

item-api/src/main/java/com/example/auction/item/api/ItemData.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Value;
+import lombok.experimental.Wither;
+
+import java.time.Duration;
+import java.util.UUID;
+
+
+/**
+ * Data is different from Info in that Info can be inferred from data while data is source of truth. In this
+ * case, ItemData is user generated and can't be restored or inferred from other sources.
```
TimMoore: I don't follow this comment... what's `Info`?

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
TimMoore: I put `creatorId` first in most of the things dealing with `itemSummaryByCreator`, because I wanted that to be the partition key.

It doesn't _really_ matter for the insert statement, but it was originally put in that order for consistency with the table definition. I don't mind changing it, but just wanted to point out that it wasn't an accident.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
```diff
@@ -71,67 +74,73 @@ public ItemServiceImpl(PersistentEntityRegistry registry, BiddingService bidding
...
         });
     }
 
     @Override
-    public ServiceCall<Item, Done> updateItem(UUID id) {
+    public ServiceCall<ItemData, Item> updateItem(UUID itemId) {
         return authenticated(userId -> item -> {
-            // todo implement
-            return null;
+            // userId may not match the item's creator Id. This is validated at Command Processing in the PersistentEntity
```
TimMoore: Is that the pattern we want? I'm more accustomed to authorization being validated at a boundary layer above the persistence layer... in other words, in the service implementation. The reason being that sometimes you want to allow for updates to be triggered by something other than a user request, and if the user is being validated within the persistence layer, you then need to come up with some kind of synthetic user ID or other workaround.

That might not be necessary in our case, but what is the pattern that we want to demonstrate?

Handling it at this layer would also allow for a more semantically meaningful HTTP status such as 403.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
```diff
@@ -71,67 +74,73 @@ public ItemServiceImpl(PersistentEntityRegistry registry, BiddingService bidding
...
+            PItemData details = Mappers.fromApi(item);
+            PItemCommand.UpdateItem updateItem = new UpdateItem(userId, details);
+            return entityRef(itemId)
+                    .ask(updateItem)
+                    .handle((pitem, updateException) -> {
+                                if (updateException != null) {
+                                    throw new TransportException(TransportErrorCode.fromHttp(409),
```
TimMoore: Some discussion:

- http://stackoverflow.com/q/3290182
- http://stackoverflow.com/q/11257470

item-impl/src/main/java/com/example/auction/item/impl/Mappers.java
```diff
@@ -0,0 +1,43 @@
+package com.example.auction.item.impl;
+
+import com.example.auction.item.api.Item;
+import com.example.auction.item.api.ItemData;
+
+/**
+ *
+ */
+public class Mappers {
```
TimMoore: This is a nice idea 😄

web-gateway/app/views/editItem.scala.html
```diff
@@ -1,47 +1,92 @@
...
+                <i class="fi-alert"></i> @errorMessage.get()
+            </div>
+        }
+
         <div class="column row">
-            @inputText(itemForm("title"))
+            @if(itemStatus.equals(ItemStatus.CREATED)){
+                @inputText(itemForm("title"))
+            } else {
+                @inputText(itemForm("title"),'readonly->'readonly)
```
TimMoore: Maybe there could be a helper function like

```scala
readOnlyIf(p: () => Boolean) = if p() Map('readonly -> 'readonly) else Map()
```

item-api/src/main/java/com/example/auction/item/api/ItemData.java
ignasi35: Info, as opposed to data, can be inferred from data. Data is the unique source of truth you must not loose. It is safe to loose info because you can recompute it from data. 
The error in the comment was to use `Info` (capitalised) because I was working on the idea to put `price`, `startInstant`, etc... fields on a separate class `ItemInfo`.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
ignasi35: If I want to review authorisation on the `ServiceImpl` I'll have to do an extra call to the `PersistentEntity`. The previous code was comparing the principal on the request with the `creatorId` in the body of the request which is not valid. I could cook a request with my credentials as principal over your Item setting on the body my creatorId.

I'm not sure this is particular example sample but I guess that demonstrating that `ServiceImpl` should avoid calling twice the same PersistentEntity as that'll lead to race conditions. I agree that since `creatorId` is never modified it'd be safe to do it in this case.

About 409/403... that's an error. When user is not allowed it should be mapped to a 403.

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
ignasi35: I don't think 400 qualifies. The client did nothing wrong. The events happened in an order that caused the update to fail. When the client sent the request, it might have still been processable but in a millisecond, the event to close the auction happened.

About 422, I think it's reserved for WebDAV.

I think a 409 applies in this case, but the response should include more info so the client can fix the request as requested in the spec: https://tools.ietf.org/html/rfc7231#page-60

item-impl/src/main/java/com/example/auction/item/impl/ItemServiceImpl.java
TimMoore: Yeah good point

PR #16, opened 2016-12-04
search service

ignasi35:
Search service provides searching features. The implementation consumes events from several upstream topics and merges data from all events into Item Documents. This implementation in particular uses ElasticSearch 5.x. 
ElasticSearch 5.x no longer supports embedded server so it has to be downloaded and started manually.


search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,273 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.junit.Assert.assertEquals;
+
+import com.example.core.*;
+import scala.concurrent.duration.FiniteDuration;
+
+// This TestCase puts a lot of pressure on elasticserach destroying and rebuilding indices on each run. Currently all
+// tests are ignored but can be run on a freshly installed elasticserach 5.0.x individually.
+// This TestCase expects an elasticsearch 5.0.x available on port 9200.
+public class SearchServiceImplTest {
```
ignasi35: Elasticsearch doesn't like me dropping and recreating indices on every test, so the three tests I wrote pass when run in isolation, they sometimes pass when run in pairs and almost never pass when run all three at once. In occasions, the same test will pass and then fail if runs are too close in time.

README.md
```diff
@@ -1,135 +1,36 @@
...
-Event emitted publically are published via a broker topic named `item-ItemEvent`.
+The auction system is the sum of 5 micro-services and a web gateway:
 
-## Commands handled
+* **[Item Service](docs/item-service.md)**: Manages the description and auction status (created, auction, completed, cancelled) of an item.
+* **[Bidding service](docs/bidding-service.md)**: Manages bids on items.
+* **[Search service](docs/search-service.md)**: Handles all item searching.
+* **[Transaction service](docs/transaction-service.md)**: Handles the transaction of negotiating delivery info and making payment of an item that has completed an auction.
+* **user-service**: a convenience service to stub user management. Don't use any code in `user-service` as reference on how to create a secure user management micro-service.
+* **web-gateway**: a [Play](https://www.playframework.com/) application providing web UI and acting as gateway to all previously described services.
```
ignasi35: I split each service docs into a separate page so the main README remains smaller and we can focus the attention to higher level details.

Each service-dedicated file is split into commands, queries, events emitted (and topic name) and events consumed. Other than those 4 section each page may add arbitrary info specific to that service.

search-impl/src/test/java/com/example/core/TopicStub.java
```diff
@@ -0,0 +1,61 @@
...
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import com.lightbend.lagom.javadsl.api.broker.Subscriber;
+import com.lightbend.lagom.javadsl.api.broker.Topic;
+
+import java.util.concurrent.CompletionStage;
+import java.util.function.Supplier;
+
+// WIP
+public class TopicStub<T> implements Topic<T> {
```
ignasi35: This was extracted from the code we worked together.

search-impl/src/test/java/com/example/core/TopicStub.java
```diff
@@ -0,0 +1,61 @@
...
+                        .run(materializer);
+                sourceActor = pair.first();
+                return pair.second();
+            }
+        };
+    }
+
+    public Supplier<ActorRef> actorSupplier() {
+        return () -> sourceActor;
+    }
```
ignasi35: Have to use a `Supplier<ActorRef>`  because of the ordering in which the `TopicStub` is created, the `atLeastOnce` method is invoked and the `ActorRef`is assigned to a local variable in the client test.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,273 @@
...
+    // -----------------------------------------------------------------
+
+    public static class BiddingStub implements BiddingService {
+        private final TopicStub<BidEvent> topicStub;
+
+        @Inject
+        public BiddingStub(Materializer materializer) {
+            topicStub = new TopicStub<>(materializer);
+            bidStub = topicStub.actorSupplier();
+        }
```
ignasi35: A Stubbed service can locally create as many stubs as Topic it will use. After creating the `TopicStub` the user may request the `actorSupplier` and assign it to a test field (see code above).

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,273 @@
...
+    @AfterClass
+    public static void afterAll() {
+        testServer.stop();
+    }
+
+    private static ServiceTest.TestServer testServer;
+    private static SearchService searchService;
+
+    private static Supplier<ActorRef> bidStub;
+    private static Supplier<ActorRef> itemStub;
```
ignasi35: This will eventually return the actor inside the `TopicStub`.

search-impl/src/test/java/com/example/core/InMemServiceLocator.java
```diff
@@ -0,0 +1,84 @@
...
+    // This constructor is exploiting a weakness in the scala-java interop: `TestServiceLocatorPort` is a scala class
+    // that's "package private" and is on an 'internal' lagom package it should not be used.
+    //
+    // http://www.scala-lang.org/files/archive/spec/2.11/05-classes-and-objects.html#private
+    //
+    @Inject
+    public InMemServiceLocator(TestServiceLocatorPort testPort) {
+        this.testPort = testPort;
+    }
+
```
ignasi35: This `InMemServiceLocator` is the hack I used to locate elasticsearch on tests. It works only because `TestServiceLocatorPort` is package private in scala (which is mapped to the java equivalent of a public class when creating the bytecode). On top of that `TestServiceLocatorPort` is in package `internal`.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,273 @@
...
+// This TestCase expects an elasticsearch 5.0.x available on port 9200.
+public class SearchServiceImplTest {
+
+    private static final Setup setup = defaultSetup()
+            .configureBuilder(b ->
+                    b.overrides(
+                            bind(BiddingService.class).to(BiddingStub.class),
+                            bind(ItemService.class).to(ItemStub.class),
+                            bind(ServiceLocator.class).to(InMemServiceLocator.class)
+                    )
```
ignasi35: Binds the `InMemServiceLocator`.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,273 @@
...
+            );
+
+    @BeforeClass
+    public static void beforeAll() {
+        testServer = ServiceTest.startServer(setup);
+
+        InMemServiceLocator serviceLocator = (InMemServiceLocator) testServer.injector().instanceOf(ServiceLocator.class);
+        serviceLocator.registerService("elastic-search", 9200);
+        serviceLocator.registerService("elastic-search-test-utils", 9200);
+
```
ignasi35: Register the services I want to locate during my test.

search-impl/src/test/java/com/example/elasticsearch/ElasticsearchTestUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    default public Descriptor descriptor() {
+        return named("elastic-search-test-utils")
+                .withCalls(
+                        Service.restCall(Method.DELETE, "/:index", this::deleteIndex)
+                )
+                .withAutoAcl(true);
+    }
+}
```
ignasi35: Convenience Service to be used in tests only that allows deleting an ES index.

docs/search-service.md
```diff
@@ -0,0 +1,20 @@
+
+# Search service
+
+Handles all item searching.
+
+The implemetation is leveraging an Elasticsearch server to store and then search Items. To access Elasticsearch this service uses the [HTTP API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs.html) so this is an example of wrapping a 3rd party HTTP API on a [Lagom Descriptor](http://www.lagomframework.com/documentation/).
```
TimMoore: typo "implemetation" --> "implementation"

TimMoore: Is "Lagom Descriptor" meant to link to the documentation index? Looks like maybe it should go to http://www.lagomframework.com/documentation/1.2.x/java/ServiceDescriptors.html

item-api/src/main/java/com/example/auction/item/api/PaginatedSequence.java
```diff
@@ -6,6 +6,7 @@
 /**
  * A partial sequence of elements, with metadata for retrieving additional pages.
  */
+// TODO: move to a common sbt project to reuse from searchApi
```
TimMoore: Are you going to do this as part of this PR?

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -34,7 +34,6 @@
     @Inject
     public ItemRepository(CassandraSession session, ReadSide readSide) {
         this.session = session;
-        session.underlying().thenAccept(s -> registerCodec(s, new EnumNameCodec<>(ItemStatus.class)));
```
TimMoore: I wasn't really sure if the constructor was the best place to do this, but I'm curious about your reason for moving it.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -168,7 +163,12 @@ public PItemEventProcessor(CassandraSession session, CassandraReadSide readSide)
             );
         }
 
+        private void registerCodec(Session session, TypeCodec<?> codec) {
+            session.getCluster().getConfiguration().getCodecRegistry().register(codec);
+        }
+
         private CompletionStage<Done> prepareStatements() {
+            session.underlying().thenAccept(s -> registerCodec(s, new EnumNameCodec<>(ItemStatus.class)));
```
TimMoore: Probably should be added into the `doAll` sequence so that the `CompletionStage` returned by this method awaits this as well.

search-impl/src/main/java/com/example/auction/search/impl/SearchServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    @Inject
+    public SearchServiceImpl(
+            Elasticsearch elasticsearch,
+            ItemService itemService,
+            BiddingService biddingService) {
+        this.elasticsearch = elasticsearch;
+
+        // TODO: use ES' _bulk API
+        itemService.itemEvents().subscribe().atLeastOnce(Flow.<ItemEvent>create().map(this::toDocument).mapAsync(1, this::store));
+        biddingService.bidEvents().subscribe().atLeastOnce(Flow.<BidEvent>create().map(this::toDocument).mapAsync(1, this::store));
```
TimMoore: I find it useful to split topic subscribers out into a separate class/Guice component. It makes things easier to test, and it's better separation of responsibilities.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
TimMoore: What if the tests could be structured so that it doesn't require destroying the index between each test method?

One option that comes to mind is to include a bit of random text in the item name and the search query for each test method.

Alternatively, if the category search were fleshed out, you could use a random category for each item and include that in the search.

README.md
```diff
@@ -1,135 +1,36 @@
...
+# INTRO
 
-System architecture:
+This is a sample app of the Lagom Framework.
 
-# Item service
+## Running: Prerequisites
 
-Manages the description and auction status (created, auction, completed, cancelled) of an item.
+You will need to download and run an Elastisearch server:
```
ignasi35: It meant doing too much sbt-ing to end up invoking `java.lang Process` and hoping the user would have `tar`, `gzip` installed and then discovering the OS... Plus potential changes on download locations of binary names in ES. The app starts without ES, and when ES is available, then you can also search.

ignasi35: You made me realise I didn't test on conductr.

item-api/src/main/java/com/example/auction/item/api/PaginatedSequence.java
ignasi35: Yes. There's some bits and pieces that could be shared amongst projects. Pagination and how it's modelled on the service end, testing tools (`Await`, `expectRethrows`,...).

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
ignasi35: Oops!

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
ignasi35: TBH, I'm not sure when's the right moment to register this codec. When running several things on my laptop I could get errors claiming this codec was not available (race conditions). Also, I'm not 100% sure it needs to be on `prepareGlobal` or if it's safe to have all nodes on our cluster register the code again and again.

search-impl/src/main/java/com/example/auction/search/impl/SearchServiceImpl.java
ignasi35: Nice! Will try that. Do you have a specific example? I think promoting this kind of patterns will be helpful.

item-api/src/main/java/com/example/auction/item/api/PaginatedSequence.java
ignasi35: Also, `CompletionStageUtils` could be moved too.

README.md
ignasi35: I will link the code from the specialised readme pages. What I'm missing on the descriptor are the  details of the events a service will be consuming.

search-impl/src/main/java/com/example/auction/search/Module.java
```diff
@@ -0,0 +1,30 @@
...
+        bindServices(serviceBinding(SearchService.class, SearchServiceImpl.class));
+
+        bindClient(BiddingService.class);
+        bindClient(ItemService.class);
+        bindClient(Elasticsearch.class);
+
+        bind(IndexedStore.class).to(IndexedStoreImpl.class);
+
+        bind(BrokerEventConsumer.class).asEagerSingleton();
+
```
ignasi35: @TimMoore Extracting the topic consumption into a separate class that's how the Module looks like. The `bind(BrokerEventConsumer.class).asEagerSingleton();` is the class that builds the stream and consumes from the broker. It doesn't implement any interface because it doesn't have any method. Everything happens on the constructor so if it can resolve it's dependencies it starts consuming.

search-impl/src/main/java/com/example/auction/search/impl/BrokerEventConsumer.java
```diff
@@ -0,0 +1,70 @@
...
+
+@Singleton
+public class BrokerEventConsumer {
+
+    private IndexedStore indexedStore;
+    private ItemService itemService;
+    private BiddingService biddingService;
+
+    @Inject
+    public BrokerEventConsumer(IndexedStore indexedStore, ItemService itemService, BiddingService biddingService) {
```
ignasi35: And here's the Class taking care of event consumption. It is now doing both topics in a single class. I don't want to strip it further because it'd take the abstraction too far.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
```diff
@@ -168,8 +168,10 @@ private void registerCodec(Session session, TypeCodec<?> codec) {
         }
 
         private CompletionStage<Done> prepareStatements() {
-            session.underlying().thenAccept(s -> registerCodec(s, new EnumNameCodec<>(ItemStatus.class)));
             return doAll(
+                    session.underlying()
+                            .thenAccept(s -> registerCodec(s, new EnumNameCodec<>(ItemStatus.class)))
+                            .thenApply(x -> Done.getInstance()),
```
TimMoore: You could also use the `accept` helper from `CompletionStageUtils` that is used elsewhere in this class. It wraps a consumer into a function that returns done. Just makes things a little tidier 😄

search-impl/src/main/java/com/example/auction/search/impl/BrokerEventConsumer.java
TimMoore: Yeah, I think both topics in one class definitely makes sense.

search-impl/src/main/java/com/example/auction/search/Module.java
TimMoore: Yeah, the more I think about it, the more I think that merging `BrokerEventConsumer` into `IndexedStoreImpl` would be an improvement. Then, the ingestion of events into the index could be completely encapsulated, and it provides the `search` method as its interface.

ignasi35: I'm not sure I'm following your reasoning. I wanted to make it obvious in some way or another that `searchserrvice-impl` consumes from two topics.
In this case in particular events incoming from the topics will go straight into the `IndexedStoreImpl` but in other services the event will act as commands for a Persistent Entity.

At the moment, the responsibilities of `BrokerEventConsumer` are building the stream and adapting the public API's of upstream services into `IndexedStore`'s API. There's no intermediate representation of the data but at least  `IndexedStore` doesn't know anything about upstream API's.

Maybe what we need is for `IndexedStoreImpl` to provide a Sink we can plug the `Source`'s into.

ignasi35: I want to review if there's any difference in running the streams on a Guice `Singleton` vs running them on a Lagom managed `Service`.

ignasi35: > Maybe what we need is for IndexedStoreImpl to provide a Sink we can plug the Source's into.

Can't do, `theTopic.subscribe().atLestOnce(x)' expects `x` to be a `Flow`.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
TimMoore: I'm not sure either

TimMoore: Are you saying that you can't search by substrings or keywords? Surely that's something Elasticsearch supports.

.sbtopts.debug
```diff
@@ -0,0 +1,6 @@
+-J-Xms1024M
+-J-Xmx4096M
+-J-Xss2M
+-J-XX:MaxMetaspaceSize=1024M
+-J-Xdebug
+-J-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005
```
TimMoore: I think it's simpler to run `sbt -jvm-debug 5005`

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
TimMoore: The codec is a client-side concern. It just handles mapping of objects into native Cassandra types. It does need to run on each node.

search-impl/src/test/java/com/example/auction/search/impl/SearchServiceImplTest.java
ignasi35: ES supports searching by substrings, `online-auction` as it is now, only can search by exact words.
The problem is that a search is a combination of several criteria and wanted to test them unitarily. I guess I'll have to include the _keyword_ criteria on all tests.

item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java
ignasi35: In that case, `prepare` is the place to be. ;-)

PR #19, opened 2017-01-02
TopicStub proposal for Lagom broker testkit.

ignasi35:
In order to test services that consume messages from a topic we could move `TopicStub` and `ProducerStub` to lagom core.

See https://github.com/lagom/lagom/issues/352


ignasi35:
> One question I have: what if you call subscribe multiple times? I think it won't work, because each subscriber will create a new actor, and only the first will be retained in the future.
> Maybe this is OK for a test implementation, in which case we should document it as a known limitation.

Good point! I'll keep it in mind (see comment at the end).

> Another limitation, which differs from the way it works in practice, is that new subscribers will only see new messages, not any that were published before the subscription began.

I just noticed TopicStub is currently built on `@BeforeAll` (with all services the server will start for the test) and reused on every test. Actually... with the current implementation the `ProducerStub` is provided only when the Flow inside `TopicStub` is already available so it's not possible to `send` anything before the subscription happened. It's an interesting feature.

I need to rethink the design considering the complete lifecycle of the TopicStub throughout the test (clean/flush the TopicStub between tests? Rebuild the TopicStub under the covers between tests?) and the feature to send events before there's a subscriber.

TimMoore:
I'm not sure that these concerns necessarily need to be solved right away or block the merge. The fact that it could be used to implement these tests proves that it's already useful, even with the limitations it currently has. Perhaps they could just be noted in the Javadoc (when ported over to Lagom), but I'll leave it up to you 😄

search-impl/src/test/java/com/example/core/TopicStub.java
```diff
@@ -49,13 +54,21 @@ public TopicStub(Materializer materializer) {
...
     }
 
-    public Supplier<ActorRef> actorSupplier() {
-        return () -> sourceActor;
+    /**
+     * @return a {@link ProducerStub} that clients (test code) can use to send messages through this {@link TopicStub}.
+     */
+    public ProducerStub<T> producerStub() {
+        return (T message) ->
+                sourceActor.thenApply(topicActor -> {
```
jroper: I think there's a race condition here.  Let's say you have the following code:

```java
ProducerStub<String> stub = topicStub.producerStub();

stub.send("1");
stub.send("2");
```

In what order will those messages arrive?  I don't think it's defined.  They are sent by successive callbacks added to the `sourceActor` future.  I don't think there's any guarantees on ordering of callbacks attached to a `CompletableFuture`, in particular, let's say the first one attaches before the future is redeemed, then the future is redeemed, and at the same time the second attaches.  It may be that it executes before the first one.

I think the best way to implement this and guarantee safety will be to create a single actor that backs the `TopicStub` stub, which buffers messages before the stream is subscribed to, and once it's subscribed, then drain the queue and send messages.  You could also support multiple calls to subscribe, having multiple actors to send to, though of course only the first actor would be guaranteed to get all the buffered messages, subsequent subscriptions will only get messages published after they've subscribed.

search-impl/src/test/java/com/example/core/TopicStub.java
```diff
@@ -49,13 +54,21 @@ public TopicStub(Materializer materializer) {
                         .via(flow)
                         .toMat(Sink.ignore(), Keep.both())
                         .run(materializer);
-                sourceActor = pair.first();
+                sourceActor.complete(pair.first());
                 return pair.second();
```
jroper:
```java
Source.actorRef(1, OverflowStrategy.fail())
```

I think the above is too low a buffer - you'd only have to send twice in a row and you'd risk the stream failing.  It's safe to set it at something quite high, maybe 1024.

search-impl/src/test/java/com/example/core/TopicStub.java
```diff
@@ -12,14 +12,19 @@
...
-import java.util.function.Supplier;
 
-// WIP
+/**
+ * Stubs a {@link Topic} to run tests without starting the complete broker infrastructure. While {@link TopicStub}
+ * should only be used inside the Stub's of your Service, you can distribute the {@link ProducerStub} returned by
+ * {@link TopicStub#producerStub()} so that clients (test code) can send messages via this stubbed topic.
+ * @param <T>
+ */
 public class TopicStub<T> implements Topic<T> {
```
jroper: I can't comment on the specific lines of code because they weren't modified in this PR, but `withGroupId` needs to return `this`, and `atMostOnceSource` needs to be implemented.  `topicId` should probably not return `null`, perhaps throw a `UnsupportedOperationException`.

search-impl/src/test/java/com/example/core/TopicStub.java
jroper: If you backed `TopicStub` with an actor that was created when it was created, then this method could be replaced with a simple `send` method that would send directly to the actor.

TimMoore: Yeah, I like that idea.

If we want to support sending messages before there are subscribers and new subscribers seeing existing messages, we could pair that actor with a list of sent messages. Then, the `send` method could both send to the actor and add to the list. The subscribe method could concatenate a source built from the list with the actor source.

ignasi35: Tim's suggestion to support `send` before subscription had my rewriting ProducerStub to contain an actor and subscribe connecting Producer's actor to the flow. This approach is cleaner.

ignasi35:
Split `TopicStub` into several classes to support:

 * receive messages if subscription happens after message was sent
 * receive messages in order (fixes race condition)
 * receive messages in order even if produced in different producers
 * deliver each message to several consumers (by means of groupId)
 * Introduces `TopicStubFactory` so that injection into `DeveloperServiceStub` is easier.

TODO: support `atMostOnceSource` in `SubscriberStub`.

tools/src/test/java/com/example/core/broker/testkit/services/DeltaDownstreamServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import org.pcollections.TreePVector;
+
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class DeltaDownstreamServiceImpl implements DeltaDownstreamService {
+
+    private final List<ReceivedMessage> receivedMessages = new ArrayList<>();
```
TimMoore: I think there could be a thread safety issue using an unsynchronized `ArrayList` here. May be better to use a volatile, mutable reference to a `TreePVector` the whole way through, or alternatively an immutable reference to a mutable `ConcurrentLinkedQueue`.

tools/src/main/java/com/example/core/broker/testkit/TopicStub.java
```diff
@@ -0,0 +1,50 @@
...
+        return new SubscriberStub<>(topicId, bufferActor, materializer);
+    }
+
+    /**
+     * @return a {@link ProducerStub} that clients (test code) can use to send messages through this {@link TopicStub}.
+     */
+    public ProducerStub<T> producerStub() {
+        return (T message) -> {
+            bufferActor.tell(message, ActorRef.noSender());
+        };
```
TimMoore: Is there a benefit of having a separate `ProducerStub` object rather than implementing `send` directly on `TopicStub`?

ignasi35: Clients only see the `send`method because `TopicStub` will not escape from `ServiceStub` which prevents clients (test code) to invoke `subscribe`.

tools/src/test/java/com/example/core/broker/testkit/services/DeltaDownstreamServiceImpl.java
ignasi35: 🙇 
My original impl was ignoring the fact that `TreePVector.from(Collection)` uses the Collection's `iterator`.

Will actually use a mutable queue in `CharlieDownstream` (a dummy service consuming from two topics).
Will use a `volatile PSequence` in `DeltaDownstream` and `FoxtrotDownstream` (since Delta and Foxtrot only consume from one topic). But I think `volatile` is unnecessary since we accept the `Service` return eventually consistent values.

tools/src/main/java/com/example/core/broker/testkit/SubscriberStub.java
```diff
@@ -36,12 +34,15 @@ public SubscriberStub(String topicId, String groupId, ActorRef topicBuffer, Mate
...
+    }
+
     @Override
     public Source<T, ?> atMostOnceSource() {
         Source<T, ActorRef> source = Source.<T>actorRef(1024, OverflowStrategy.fail());
-        Pair<ActorRef, CompletionStage<Done>> pair = source.toMat(Sink.ignore(), Keep.both()).run(materializer);
-        topicBuffer.tell(new SubscribeToBuffer(groupId, pair.first()), ActorRef.noSender());
-        return source;
+        return source.prependMat(Source.empty(), this::registerActor);
     }
```
TimMoore: I believe that `atLeastOnce` can now be implemented in terms of `atMostOnceSource`:

```java
    public CompletionStage<Done> atLeastOnce(Flow<T, Done, ?> flow) {
        return atMostOnceSource()
                .via(flow)
                .toMat(Sink.ignore(), Keep.right())
                .run(materializer);
    }
```

ignasi35: Brilliant! +1

ignasi35:
@jroper , @TimMoore I'm about to start moving this into core. But before that I would like you opinion on `TopicStubFactory`. Now it provides `TopicStub`s for anything that's requested but it could generate them from a Descriptor. That is more maintainable but it requires some reflection. 

Before jumping head-first into that alternative I am considering completing the task with the current approach but marking the `TopicStub`, `TopicStubFactory` and `ProducerStub` as experimental.

TimMoore:
I see what you mean... I think it's nice to be able to create arbitrary ones, rather than coupling it to a Descriptor, but it would be nice to offer reading a Descriptor as an option so that you don't need to expose the topic name as a constant this way.

I'm 👍 for completing it as is and enhancing more in separate PRs.

tools/src/main/java/com/example/core/broker/testkit/TopicStub.java
jroper: I think the point here is why do we need a `ProducerStub` at all?  Why not just have a `send` method?  At the moment you've got this boiler plate in every test:

```java
topicStub = topicFactory.forId(BidEvent.class, TOPIC_ID);
bidTopicStub = topicStub.producerStub();

...
bidTopicStub.send(...);
```

Why not just do:

```java
topicStub = topicFactory.forId(BidEvent.class, TOPIC_ID);

...
topicStub.send(...);
```

I think the reasons for making it this way originally now no longer exist.

tools/src/main/java/com/example/core/broker/testkit/SubscribeToBuffer.java
```diff
@@ -0,0 +1,47 @@
+package com.example.core.broker.testkit;
+
+import akka.actor.ActorRef;
+
+public class SubscribeToBuffer {
```
jroper: Style note, this is part of the `TopicBufferActor` protocol, Akka convention for Java code is to put all classes that make up the protocol of an actor as inner classes of a class called `<ActorClass>Protocol`, ie in this case `TopicBufferActorProtocol`.  Also, this class should be package private.

tools/src/main/java/com/example/core/broker/testkit/TopicStub.java
ignasi35: The code, as is, allows creating many producers for a single topic like you'd get in a real brokering environment. Having many producers for a single topic in test wouldn't make much  sense though, so I agree with you that my original motivation is gone.

In general, I am trying to avoid adding methods to an impl of `Topic`. I want `TopicStub` to be just an in-memory `Topic`.
Also, I don't like that this design has 3 public classes either (`ProducerStub`, `TopicStub` and `TopicFactoryStub`). I would like to reduce it to a minimum and I'm trying to hide `TopicStub` completely so the user only sees `ProducerStub` and a `SomethingFactory`:

```
        MyServiceStub(TopicStubFactory topicFactory) {
            myProducerStub = topicFactory.producer(MyEvent.class, TOPIC_ID);
        }
        @Override
        public Topic<MyEvent> messageTopic() {
            return myProducerStub.topic();
        }
```
and then use it for sending:
```
       myProducerStub.send(MyEvent.of("the message"));
```


Probably I should rename the `TopicStubFactory` to `ProducerFactory` or `TopicProducerTestkit` but there's an (unrelated) `TopicProducer` in production code and I don't want to mix things up.

tools/src/main/java/com/example/core/broker/testkit/SubscriberStub.java
```diff
@@ -0,0 +1,58 @@
...
+        this.topicId = topicId;
+        this.groupId = groupId;
+        this.topicBuffer = topicBuffer;
+        this.materializer = materializer;
+    }
+
+    @Override
+    public Subscriber<T> withGroupId(String groupId) throws IllegalArgumentException {
+        return new SubscriberStub<>(topicId, groupId, topicBuffer, materializer);
+    }
```
ignasi35: Implements support for consumer groups via `groupId` so that two services (in the same microservice) consuming from the same topic don't interfere. I had a `security-service-impl` in mind where there's a `CredentialsServiceImpl` and a `TokenServiceImpl` both consuming from a fraud topic.

tools/src/main/java/com/example/core/broker/testkit/SubscribeToBuffer.java
ignasi35: Thanks, will rewrite.
I was aware of the scala convention but forgot to check if there was one for java too. When porting to core I rewrote almost all in scala but we'll keep this here until bumping online-auction to 1.3 anyway so it'd better be convention-compliant.

ignasi35:
This branch is now ready for merge when a new 1.3.x release is prepared.

ignasi35:
With 1.3.0-RC1 this branch can be updated and cleaned up (remove `ProducerStub`and related classes to use those from lagom). 

Don't merge until 1.3.0 is out.

ignasi35:
Superseded by https://github.com/lagom/online-auction-java/issues/34

PR #24, opened 2017-01-20
Make the circuit breakers more tolerant of slow-to-start services

tommpy:
On my machine the downstream services take a few seconds to respond to the very first requests which trips the default config for the circuit breakers.  This seems to work around the problem so now the frontend responds as soon as runAll finishes.

In debugging this, one suggestion was that sbt could have an effect, so I updated it to avoid any confusion.

### DIFF
--- a/web-gateway/conf/application.conf
+++ b/web-gateway/conf/application.conf
@@ -1,7 +1,3 @@
-
-
-lagom.circuit-breaker.default.call-timeout = 5s
-
 play.crypto.secret = "somesecret"
 
 lagom.play {
@@ -11,13 +7,4 @@ lagom.play {
       path-regex = "(?!/api/).*"
     }
   ]
-}
-
-lagom.circuit-breaker {
-  default {
-    enabled = on
-    max-failures = 10
-    call-timeout = 10s
-    reset-timeout = 15s
-  }
 }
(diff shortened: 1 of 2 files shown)


benmccann:
I wonder if this affects other Lagom projects as well. Maybe we should change the Lagom defaults as a more comprehensive way to address this issue?

tommpy:
True, I guess I wanted to dig a bit deeper in to why they're taking more that 3 seconds to respond to the first request.

web-gateway/conf/application.conf
```diff
@@ -11,4 +11,13 @@ lagom.play {
...
   ]
+}
+
+lagom.circuit-breaker {
+  default {
+    enabled = on
+    max-failures = 10
+    call-timeout = 10s
+    reset-timeout = 15s
+  }
```
ignasi35: Hi @tommpy, the values you suggest for the `lagom.circuit-breaker.default` are equal to Lagom's defaults: https://github.com/lagom/lagom/blob/f62df3283a3006b2c1bf13c4ca07e53acce277b8/service/core/client/src/main/resources/reference.conf#L6-L24. 

I think the solution to slow downstream services bootup could be to remove the setting on line 3 which makes the circuit breakers act a bit more aggressively than defaults.

PR #25, opened 2017-01-24
Create one PersistentEntitytestDriver per @Test.

ignasi35:
The test execution in ~sbt~ _junit_ is creating one instance of the test class for each @Test method which causes the `PersistentEntitytestDriver driver` instance to be different on each @Test. 

The test, as previously written gave the impression that PersistentEntitytestDriver provided by Lagom is stateless and that a single instance can be reused between tests which is not true.


TimMoore:
As discussed yesterday, this isn't strictly necessary, because most JUnit runners also create an instance per test, but may be a good idea if the `PersistentEntityTestDriver` reference retains a lot of memory (in which case, it's a good idea to clear it to `null` in an `@After` block, too).

PR #37, opened 2017-02-23
Small fix for bug #30

### LINKED ISSUE #30: It's impossible to create an auction with a currency of Japanese Yen
When you create an auction in Japanese Yen, it's impossible to enter a valid increment, if it's less than 50, you get an error of invalid.step, if it's more than 50, you get an error of invalid.increment.

yg-apaza:
Japanese currency doesn't use decimal places (currency.getDefaultFractionDigits() returns 0). In this case the step should be 1.
Fixes #30

### DIFF
--- a/web-gateway/app/controllers/Currency.java
+++ b/web-gateway/app/controllers/Currency.java
@@ -77,4 +77,9 @@ public enum Currency {
         int price = toPriceUnits(value);
         return price % step == 0;
     }
+    
+    public boolean isValidIncrement(double value) {
+    	int price = toPriceUnits(value);
+    	return price > 0 && price < 100 * this.step;
+    }
 }
--- a/web-gateway/app/controllers/ItemForm.java
+++ b/web-gateway/app/controllers/ItemForm.java
@@ -31,16 +31,13 @@ public class ItemForm {
         List<ValidationError> errors = new ArrayList<>();
         
         try {
-        	Currency c = Currency.valueOf(currency);
+            Currency c = Currency.valueOf(currency);
             if (!c.isValidStep(increment.doubleValue())) {
                 errors.add(new ValidationError("increment", "invalid.step"));
             }
-
             if (!c.isValidStep(reserve.doubleValue())) {
                 errors.add(new ValidationError("reserve", "invalid.step"));
             }
-            // Make sure that the increment and reserve are multiples of 50c - in a real app, this would be more complex
-            // and based on currency specific rules, for now we'll assume currencies that have cents.
             if (!c.isValidIncrement(increment.doubleValue())) {
                 errors.add(new ValidationError("increment", "invalid.increment"));
             }
--- a/web-gateway/app/controllers/ItemForm.java
+++ b/web-gateway/app/controllers/ItemForm.java
@@ -31,8 +31,7 @@ public class ItemForm {
         List<ValidationError> errors = new ArrayList<>();
         
         try {
-            Currency c = Currency.valueOf(currency);
-
+        	Currency c = Currency.valueOf(currency);
             if (!c.isValidStep(increment.doubleValue())) {
                 errors.add(new ValidationError("increment", "invalid.step"));
             }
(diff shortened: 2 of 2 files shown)


jroper:
While a step of 1 works - 1 Yen is only 1c, I think we want the step to be 50 Yen, so that you don't have auctions where people are just putting 1 Yen increments in.

The problem I think is in the validation logic, in the ItemForm, which assumes US currency.  The maximum value for the increment should probably be something like 100 times the step, so in US, where the step is 50c, that means $50, while if in Yen, it's 50 Yen, then that would be 5000 Yen as a maximum increment.

yg-apaza:
Oh, I see, can I create another method _isValidIncrement()_ in Currency.java? I think I would need to use currency.getDefaultFractionDigits() to get 10^(2 - n) * step

web-gateway/app/controllers/ItemForm.java
```diff
@@ -29,10 +29,8 @@
 
     public List<ValidationError> validate() {
         List<ValidationError> errors = new ArrayList<>();
-        
+        Currency c = Currency.valueOf(currency);
         try {
-            Currency c = Currency.valueOf(currency);
-
             if (!c.isValidStep(increment.doubleValue())) {
```
ignasi35: I think the `Currency c = Currency.valueOf(currency);` statement should be inside the `try{} catch{}` block because providing the currency `ASDF` will cause to throw an `IllegalArgumentException`.

That will require your other change (below) to be moved inside this `try{} catch` .

web-gateway/app/controllers/Currency.java
```diff
@@ -77,4 +77,9 @@ public boolean isValidStep(double value) {
         int price = toPriceUnits(value);
         return price % step == 0;
     }
+    
+    public boolean isValidIncrement(double value) {
+    	int price = (int) (value * 2);
+    	return price > 0 && price < 100 * Math.pow(10, 2 - currency.getDefaultFractionDigits());
+    }
```
jroper: How about:

```java
int price = toPriceUnits(value);
return price > 0 && price < 100 * step;
```

In US dollars, this makes it 0 < price < $50, and in Yen this makes it 0 < price < 5000.

yg-apaza:
Thanks @jroper , I squashed the commits.

web-gateway/app/controllers/ItemForm.java
```diff
@@ -31,30 +31,21 @@
         List<ValidationError> errors = new ArrayList<>();
         
         try {
-            Currency c = Currency.valueOf(currency);
-
+        	Currency c = Currency.valueOf(currency);
```
TimMoore: Please make sure you're using only space characters to indent, not tabs. You can probably set up your editor or IDE to do this automatically... let us know what tool you're using if you'd like a pointer on how to set it up.

web-gateway/app/controllers/ItemForm.java
```diff
@@ -31,30 +31,21 @@
...
-
-        // Make sure that the increment and reserve are multiples of 50c - in a real app, this would be more complex
-        // and based on currency specific rules, for now we'll assume currencies that have cents.
-        try {
-            int incrementInt = increment.multiply(BigDecimal.valueOf(2)).intValueExact();
-            if (incrementInt <= 0) {
-                errors.add(new ValidationError("increment", "invalid.increment"));
-            } else if (incrementInt >= 100) {
+            // Make sure that the increment and reserve are multiples of 50c - in a real app, this would be more complex
+            // and based on currency specific rules, for now we'll assume currencies that have cents.
```
TimMoore: I think this comment is no longer accurate. With this change, it's not assuming cents, and the rule is currency-specific (though still fairly simple). I think the comment can just be deleted.

web-gateway/app/controllers/ItemForm.java
yg-apaza: I'm using Eclipse, how can I configure it?

TimMoore: @yg-apaza this answer on Stack Overflow covers it well http://stackoverflow.com/a/408449/29470

PR #46, opened 2017-02-27
Ensure @Value annotation on immutable classes

### LINKED ISSUE #9: Ensure all value classes implement equals/hashCode/toString
See https://github.com/lagom/online-auction-java/pull/6#discussion_r86748638

yg-apaza:
Fixes #9

### DIFF
--- a/bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionState.java
+++ b/bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionState.java
@@ -43,7 +43,7 @@ public final class AuctionState implements Jsonable {
     public static AuctionState start(Auction auction) {
         return new AuctionState(Optional.of(auction), AuctionStatus.UNDER_AUCTION, TreePVector.empty());
     }
-    
+
     public AuctionState withStatus(AuctionStatus status) {
         return new AuctionState(auction, status, biddingHistory);
     }
--- a/bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionState.java
+++ b/bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionState.java
@@ -1,6 +1,8 @@
 package com.example.auction.bidding.impl;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.Value;
+import lombok.experimental.Wither;
 import com.lightbend.lagom.serialization.Jsonable;
 import org.pcollections.PSequence;
 import org.pcollections.TreePVector;
@@ -10,6 +12,8 @@ import java.util.Optional;
 /**
  * The auction state.
  */
+@Value
+@Wither
 public final class AuctionState implements Jsonable {
 
     private static final long serialVersionUID = 1L;
@@ -42,10 +46,6 @@ public final class AuctionState implements Jsonable {
         return new AuctionState(Optional.of(auction), AuctionStatus.UNDER_AUCTION, TreePVector.empty());
     }
 
-    public AuctionState withStatus(AuctionStatus status) {
-        return new AuctionState(auction, status, biddingHistory);
-    }
-
     public AuctionState bid(Bid bid) {
         if (lastBid().filter(b -> b.getBidder().equals(bid.getBidder())).isPresent()) {
             // Current bidder has updated their bid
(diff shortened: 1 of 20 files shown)


bidding-impl/src/main/java/com/example/auction/bidding/impl/AuctionState.java
```diff
@@ -10,6 +12,8 @@
 /**
  * The auction state.
  */
+@Value
+@Wither
```
ignasi35: I think this `@Wither` added more methods than there where originally. Before this PR, only `withStatus` was available.

search-impl/src/main/java/com/example/elasticsearch/Query.java
```diff
@@ -2,10 +2,12 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
 
 /**
  *
  */
+@Value
 public class Query {
```
ignasi35: The classes modelling the API to ElasticSearch don't always need the `@Value` annotation.

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionEvent.java
```diff
@@ -1,11 +1,13 @@
 package com.example.auction.transaction.api;
 
 import java.util.UUID;
+import lombok.Value;
 
 public abstract class TransactionEvent {
```
ignasi35: All the code in `transaction-xxx` is WIP and it'd be good to not add more complexity on it until we have a clear idea what is it we want to do with it.

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionMessage.java
```diff
@@ -2,7 +2,9 @@
 
 import java.time.Instant;
 import java.util.UUID;
+import lombok.Value;
 
+@Value
 public final class TransactionMessage {
```
ignasi35: See comment above

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionInfo.java
```diff
@@ -2,7 +2,9 @@
 
 import com.example.auction.item.api.Item;
 import org.pcollections.PSequence;
+import lombok.Value;
 
+@Value
 public final class TransactionInfo {
```
ignasi35: See comment above

ignasi35:
This clashes with #52

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionEvent.java
```diff
@@ -5,7 +5,7 @@
...
 
     private TransactionEvent() {}
-
+
```
TimMoore: It looks like some extraneous whitespace changes snuck in.

Here are some instructions for configuring Eclipse to strip trailing whitespace on save: http://stackoverflow.com/a/14178860

user-impl/src/main/java/com/example/auction/user/impl/UserCommand.java
```diff
@@ -4,44 +4,19 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.lightbend.lagom.javadsl.persistence.PersistentEntity;
 import com.lightbend.lagom.serialization.Jsonable;
+import lombok.Value;
 
 import java.util.Optional;
 
 public interface UserCommand extends Jsonable {
+	@Value
```
TimMoore: It looks like some files are indenting with tab characters instead of space characters.

Here are instructions for configuring Eclipse to use spaces instead of tabs: http://stackoverflow.com/a/408449

yg-apaza:
:sweat_smile: I forgot to configure properly formatting in Eclipse, now I fixed it.

TimMoore:
@ignasi35 I'll let you have one last look before merging

build.sbt
```diff
@@ -136,7 +142,10 @@ lazy val userApi = (project in file("user-api"))
   .settings(commonSettings: _*)
   .settings(
     version := "1.0-SNAPSHOT",
-    libraryDependencies += lagomJavadslApi
+    libraryDependencies ++= Seq(
+      lagomJavadslApi,
+      lombok
```
ignasi35: I think this dependency is not used ATM.

PR #53, opened 2017-03-08
Add elasticsearch in travis

### LINKED ISSUE #48: Enable ElasticSearch tests on Travis
Use Travis CI's support for databases and ScalaTest tagging to run `search-impl` integration tests with ElasticSearch.

A possible issue will be the version miss-match between Travis options and our code/API client.

See also https://github.com/lagom/lagom/issues/569

lakhina:
- Use tar approach as deb installation required sudo access but containers do not provide sudo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,16 @@ language: scala
 sudo: false
 jdk:
   - oraclejdk8
+before_install:
+  - location=`pwd`
+  - cd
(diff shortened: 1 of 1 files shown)


ignasi35:
Hi @lakhina, this PR doesn't fully solve #48 since it only creates the Travis setup to have an elastisearch instance available. This PR is not enabling the tests on `search-impl`. I edited your comment to remove the first bullet and we can work on the second part of the issue in a separate PR.

PR #61, opened 2017-03-15
Add conditional test run on elastic search

### LINKED ISSUE #48: Enable ElasticSearch tests on Travis

lakhina:
fixes #48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: scala
-sudo: false
+sudo: required
 jdk:
   - oraclejdk8
 before_install:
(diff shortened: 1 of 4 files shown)


ignasi35:
This looks great!

I will merge as is and create 2 issues: 

 * track potential changes on installing ES via .deb files. IIUC there was some work on that front already but it happened before using `sudo: required`.
 * restore `sbt +test`. I'm not sure we need to run the tests for multiple scala versions but want to have the reasons of that change tracked.

Issue #56, opened 2017-03-13
INSTRUCTIONS: create users


ignasi35:
Add inline instructions to create several users

mehulsr:
Hi, my PR for this issue was merged. It is still open; does it need enhancement?

TimMoore:
@mehulsr this is done now. Thanks for pointing it out.

For future reference, GitHub has a convention where if you add "Fixes #56" (or whatever the issue number is) to the description of the pull request, it will close the issue automatically when the pull request is merged. So it's better to use this convention rather than putting the issue number in the title of the pull request, where GitHub doesn't look for it.

Here's GitHub's documentation on how this works: https://help.github.com/articles/closing-issues-via-commit-messages/

mehulsr:
@TimMoore Okay, will keep this in mind for the next PR. Thanks!

PR #64, opened 2017-03-16
Issue #56: Added inline instruction for creating new user.

mehulsr:
Hi, I am Mehul Smriti Raje, intending to apply for Outreachy 14 with Lagom. 

Adding fix for #56  
Is this the correct way to add the inline instruction or do we want a placeholder text?

### DIFF
--- a/web-gateway/app/views/createUser.scala.html
+++ b/web-gateway/app/views/createUser.scala.html
@@ -9,7 +9,7 @@
     @foundationForm(createUser, routes.Main.createUser()) {
         <div class="row">
             <div class="medium-6 columns">
-              <p>Enter username in the box below and click on 'Create User' to create a new user profile.</p>
+              <p>@message("createUserInstructions")</p>
                 @inputText(createUser("name"))
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


mehulsr:
I have signed the Lightbend and Scala CLA.

web-gateway/app/views/createUser.scala.html
```diff
@@ -9,9 +9,10 @@
     @foundationForm(createUser, routes.Main.createUser()) {
         <div class="row">
             <div class="medium-6 columns">
+              <p>Enter username in the box below and click on 'Create User' to create a new user profile.</p>
```
jroper: Could you extract this message into the `web-gateway/conf/messages` file? Perhaps call it `createUserInstructions`. Also perhaps not necessary to instruct users to click create user, but rather just state what this form is for and what it will do, maybe "Here you can create a new user who can participate in and create auctions."

mehulsr: Hi @jroper I have made the changes in a new commit.

PR #65, opened 2017-03-17
Use ES deb installers in Travis

### LINKED ISSUE #63: Use .deb installers in Travis
After #61 was merged the jobs in Travis are run using `sudo: required`. This takes the build out of a container into a VM where `sudo` can be used.

Without that restriction it may be possible to install elasticsearch by other means.

There may still be a restriction wrt installing the exact version we need to run the tests though.

lakhina:
Fixes #63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,16 @@ language: scala
 sudo: required
 jdk:
   - oraclejdk8
+addons:
+  apt:
+    packages:
+      - oracle-java8-set-default # Installing Elasticsearch from .deb causes default JDK in the system to fallback. This is required to ensure that 'java' (and other JDK commands) is linked to those of oraclejdk8
 before_install:
-  - location=`pwd`
-  - cd
-  - curl -L -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.2.tar.gz
-  - tar -xvf elasticsearch-5.0.2.tar.gz
-  - cd elasticsearch-5.0.2/bin
-  - ./elasticsearch -d
+  - curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.2.deb
+  - sudo dpkg -i --force-confnew elasticsearch-5.0.2.deb
+  - sudo service elasticsearch restart
+before_script:
   - sleep 20
-  - cd $location
 script:
   - sbt -DenableElasticsearch=true test
 cache:


.travis.yml
```diff
@@ -2,15 +2,16 @@ language: scala
 sudo: required
 jdk:
   - oraclejdk8
+addons:
+  apt:
+    packages:
+      - oracle-java8-set-default
```
ignasi35: Hmm, this looks weird, I'm guessing there's competing JDK's caused by `.deb` dependencies. Do we have some docs pointing to this approach?

ignasi35:
I'd like a second look from @TimMoore

.travis.yml
lakhina: Here  is the link of the doc https://www.peterbe.com/plog/elasticsearch-5-in-travis-ci .

lakhina: Yes it was failing with error with Java incompatibility 52. I tested the build with Travis on my fork and cured it before sending PR. Should I point log here?

TimMoore: Yes, I'd like to look at the log please.

lakhina: @TimMoore Switches to `jdk: - oraclejdk8` [here](https://travis-ci.org/lakhina/online-auction-java/builds/212837538#L139)
Still failing at [here](https://travis-ci.org/lakhina/online-auction-java/builds/212837538#L192)

TimMoore: OK that's interesting... I wonder why that happens.

Thanks for the extra detail... as I mentioned before, I think it would be great to add this extra context into a comment in `.travis.yml` file for the benefit of future contributors (and our future selves!)

lakhina:
@TimMoore Please take a look on it now and let me know if inline comment or anything else need to be updated

.travis.yml
```diff
@@ -2,15 +2,16 @@ language: scala
 sudo: required
 jdk:
   - oraclejdk8
+addons:
+  apt:
+    packages:
+      - oracle-java8-set-default # Important to add for removing Unsupported major.minor version 52.0 error
```
ignasi35: How about: "Installing Elasticsearch from .deb causes default JDK in the system to fallback. This is required to ensure that 'java' (and other JDK commands) is linked to those of oraclejdk8"

lakhina:
@ignasi35 I have amended the change in the inline comment.Please review and let me know if any other changes are required.

TimMoore:
Thanks, @lakhina!

PR #70, opened 2017-03-24
Added flag for inline instructions

### LINKED ISSUE #60: Inline instructions
Add a feature-flag (enabled by default) to display inline instructions on the user's UI. This means adding text snippets in the rendered HTML so that an eager user gets info in the actual UI when using the app.

Having inline instructions should help a new user get a better understanding of what can be done with the app once started. The feature flag will hide the inline instructions and is meant to remove noise if/when demoing the app.

This feature flag will control the data added in #59  (wip #68 ) , #58(done), #57(done) and #56 (done).

The existence of this inline help should be introduced in the README.md.
The steps to disable the feature should be documented in the repo README.md.

mehulsr:
Fixes #60

### DIFF
--- a/web-gateway/app/controllers/ItemController.java
+++ b/web-gateway/app/controllers/ItemController.java
@@ -32,8 +32,7 @@ public class ItemController extends AbstractController {
     private final ItemService itemService;
     private final BiddingService bidService;
 
-    private Configuration config;
-    private Boolean showInlineInstruction;
+    private final Boolean showInlineInstruction;
 
     @Inject
     public ItemController(Configuration config, MessagesApi messagesApi, UserService userService, FormFactory formFactory,
@@ -43,7 +42,6 @@ public class ItemController extends AbstractController {
         this.itemService = itemService;
         this.bidService = bidService;
 
-        this.config = config;
         showInlineInstruction = config.getBoolean("play.instruction.show");
     }
 
@@ -114,6 +112,7 @@ public class ItemController extends AbstractController {
 
                                         return ok(
                                                 views.html.editItem.render(
+                                                        showInlineInstruction,
                                                         item.getId(),
                                                         formFactory.form(ItemForm.class).fill(itemForm),
                                                         item.getStatus(),
@@ -137,7 +136,7 @@ public class ItemController extends AbstractController {
             ItemStatus itemStatus = ItemStatus.valueOf(itemStatusStr);
             if (form.hasErrors()) {
                 return loadNav(user).thenApply(nav ->
-                        ok(views.html.editItem.render(itemId, form, itemStatus, Optional.empty(), nav))
+                        ok(views.html.editItem.render(showInlineInstruction, itemId, form, itemStatus, Optional.empty(), nav))
                 );
             } else {
                 ItemData payload = fromForm(form.get());
@@ -152,7 +151,7 @@ public class ItemController extends AbstractController {
                             } else {
                                 String msg = ((TransportException) exception.getCause()).exceptionMessage().detail();
                                 return loadNav(user).thenApply(nav -> ok(
-                                        editItem.render(itemId, form, itemStatus, Optional.of(msg), nav)));
+                                        editItem.render(showInlineInstruction, itemId, form, itemStatus, Optional.of(msg), nav)));
                             }
                         }).thenCompose((x) -> x);
             }
@@ -214,8 +213,7 @@ public class ItemController extends AbstractController {
 
                 Optional<BidResult> bidResult = loadBidResult(ctx.flash());
 
-                return ok(views.html.item.render(item, bidForm, anonymizeBids(user, currency, bidHistory), user,
-                        currency, seller, winner, currentBidMaximum, bidResult, nav));
+                return ok(views.html.item.render(showInlineInstruction, item, bidForm, anonymizeBids(user, currency, bidHistory), user, currency, seller, winner, currentBidMaximum, bidResult, nav));
             });
         }));
     }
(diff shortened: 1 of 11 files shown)


mehulsr:
@ignasi35 @TimMoore I've only added the flag for instructions to create users and create items. Is this the correct way to add feature flags?

web-gateway/app/controllers/Main.java
```diff
@@ -17,6 +17,8 @@
     private final UserService userService;
     private final FormFactory formFactory;
 
+    public static boolean showInlineInstruction = true;
```
ignasi35: Hi @mehulsr, rather than using a `public static` variable, prefer using a property we can setup on `application.conf` and provide a default value via a `reference.conf`. Then, in the code, we can read the value using the `Configuration` instance (we can get it injected into each `Controller`).

I'll let you work out how to pass the value from the `Controller` to the view. ;-)

mehulsr: Sure, on it :)

TimMoore:
Hi @mehulsr are you still working on this? Thanks!

mehulsr:
@TimMoore  Yes, I am.
I had the following question:
1. I am trying to inject the configuration value in the controller using this
`    private Configuration config; `
`    @Inject
    public CreateUserForm(Configuration config){
        this.config = config;
    }`
This gives me an error stating that a default constructor is also required. Is there a change required in the way this controller object needs to instantiated?

2. I found that values can be passed from the controller to view using Result objects. I am trying to use the following:
` private boolean showInlineInstructions;`
`    public Result getShowInstructions(){
        return ok(showInlineInstruction);
    }`
How do I now check this value in the view to see if it's true or false?

TimMoore:
Thanks for your reply.

1. `CreateUserForm` isn't actually a controller, it's a [form object](https://playframework.com/documentation/2.5.x/JavaForms). I know it's in the `controllers` directory, but this also contains other classes that are _used by_ the controllers. [`ProfileController`](https://github.com/lagom/online-auction-java/blob/master/web-gateway/app/controllers/ProfileController.java) is the class you want to modify for the user page.
2. `Result` is used to create responses to web requests, not to pass values to the view. Your example actually completes the request with the value of `showInlineInstruction`.

    Play's template language ([Twirl](https://playframework.com/documentation/2.5.x/JavaTemplates)) declares [Template Parameters](https://playframework.com/documentation/2.5.x/JavaTemplates#Template-parameters) at the top of each file. You can look at the parameters that [`editItem.scala.html`](https://github.com/lagom/online-auction-java/blob/master/web-gateway/app/views/editItem.scala.html#L8) accepts for example. Then, the controller passes these directly when rendering the template, as you can see in [`ItemController.editItemForm`](https://github.com/lagom/online-auction-java/blob/master/web-gateway/app/controllers/ItemController.java#L108-L113).

I hope that gets you going in the right direction. Please let us know if you get stuck on anything else.

mehulsr:
@TimMoore 
Thanks for the comment! Please review the new commit and let me know if I got it right this time. I'll update the other files then.
Also, how would a `reference.conf` be useful in this case? @ignasi35

web-gateway/app/controllers/ItemController.java
```diff
@@ -30,19 +32,25 @@
     private final ItemService itemService;
     private final BiddingService bidService;
 
+    private Configuration config;
```
TimMoore: It isn't necessary to store the configuration in an instance variable, because it's only used inside the constructor.

web-gateway/app/controllers/Main.java
```diff
@@ -17,11 +18,19 @@
     private final UserService userService;
     private final FormFactory formFactory;
 
+    //public static boolean showInlineInstruction = true;
```
TimMoore: Please remove the commented code.

web-gateway/app/controllers/Main.java
```diff
@@ -17,11 +18,19 @@
...
+
+    private Configuration config;
```
TimMoore: Same comment as above: if you don't use config outside of the constructor, it isn't necessary to create an instance variable.

web-gateway/app/controllers/ItemController.java
```diff
@@ -30,19 +32,25 @@
...
+    private Boolean showInlineInstruction;
```
ignasi35: Prefer using `final`.

ignasi35:
> Also, how would a reference.conf be useful in this case? @ignasi35

I think, a user that removes the property from `application.conf` would end up with a `null` values instead of a default `false`. But I'm not sure really. I'd like the application to default to `false` if the property is not present and I mentioned a possible implementation instead of the actual wished behavior.

mehulsr:
I have corrected the errors pointed above. I have also added the flags for the remaining inline instructions. Please review.

web-gateway/conf/application.conf
```diff
@@ -1,2 +1,2 @@
 play.crypto.secret = "somesecret"
-
+play.instruction.show = true
```
ignasi35: For a future improvement: we could use a different namespace since `play.xyz` is meant for play, while this setting is specific to `online-auction` app.

mehulsr:
Hey @TimMoore @ignasi35 
I'd like to make these improvements. Do I open a new PR or re-open this one?

ignasi35:
HI @mehulsr this PR has already been merged, so I'd recommend you update your local `master` to `origin/master` and create a new branch and a new PR. ;-)

mehulsr:
Okay sure 👍

PR #77, opened 2017-04-07
Modified instructions in message.config

mehulsr:
Changed inline instructions to create users and edit items in message.config to match naming conventions. Updated the references to these instructions.


ignasi35:
👍

Issue #21, opened 2017-01-17
Provide a Maven example


TimMoore:
We discussed converting this project to Maven as an example for Maven/Java users.

Our current thinking is to have both `build.sbt` and `pom.xml` files on the `master` branch, and then to also have long lived `sbt` and `maven` branches that strip the opposite build files out.

lakhina:
I am working on this maven port. But having issue in replacing `enablePlugins(PlayJava && LagomPlay)` `enablePlugins(LagomJava)` 

Where can I find these plug-ins for maven?

TimMoore:
@lakhina in Maven, there is just one `lagom-maven-plugin`, but it has different options for enabling Play and Lagom projects.

Probably the easiest way to see how to configure it is to create a new Maven project using the archetype to see how that's configured.

http://www.lagomframework.com/documentation/1.3.x/java/GettingStartedMaven.html

There are some more docs on Maven builds at http://www.lagomframework.com/documentation/1.3.x/java/LagomBuild.html

However, as someone pointed out on Gitter, Play does not yet have official support for Maven, so resolving this issue may require integrating another, third-party Maven plugin for Play.

ignasi35:
There are other Lagom examples based in Maven:

 * https://github.com/lagom/lagom-java-maven-chirper-example
 * https://github.com/lagom/ibm-integration-examples
 * https://github.com/lagom/lagom-recipes

The ongoing efforts to make this project a hybrid sbt/mvn are abandoned. I'm closing this issue.

Issue #32, opened 2017-02-20
Routes collision in ConductR


ignasi35:
When deploying the application in ConductR it deems unusable due to a route collision.

The collision is for the route:

` POST /user`

that should only be resolved by `webGateway` as described in https://github.com/lagom/online-auction-java/blob/master/web-gateway/conf/routes#L6

but the default `bundle.conf` for `users-impl` is configuring the service as:

```cat ./user-impl/target/bundle/bundle/tmp/bundle.conf 
version              = "1"
name                 = "userimpl"
compatibilityVersion = "1"
system               = "userimpl"
systemVersion        = "1"
nrOfCpus             = 0.1
memory               = 402653184
diskSpace            = 200000000
roles                = ["web"]
components = {
  userimpl = {
    description      = "userImpl"
    file-system-type = "universal"
    start-command    = ["userimpl/bin/userimpl", "-J-Xms134217728", "-J-Xmx134217728", "-Dhttp.address=$USER_BIND_IP", "-Dhttp.port=$USER_BIND_PORT", "-Dplay.crypto.secret=75736572496d706c"]
    endpoints = {
      "user" = {
        bind-protocol = "http"
        bind-port     = 0
        services      = ["http://:9000/user"]  <-- I suspect this is the cause of the collision.
      },
      "akka-remote" = {
        bind-protocol = "tcp"
        bind-port     = 0
        services      = []
      }
    }
  }
}
```

*UPDATE*: similar collisions with `/item` and `/search`.

markusjura:
Having a Lagom Descriptor called `user` and another service that is using the `/user` route is not possible.

Under ConductR, each Lagom service is accessible under `http://:9000/DESCRIPTOR_NAME` where `DESCRIPTOR_NAME` is the name of the Descriptor: https://github.com/lagom/online-auction-java/blob/master/user-api/src/main/java/com/example/auction/user/api/UserService.java#L25

This suffix is necessary so that every service can use the same port 9000.

So you either rename the Descriptor name of the `user-api` service or the route of `webGateway`.

ignasi35:
> So you either rename the Descriptor name of the user-api service or ...

Exactly what I did. ;-)

ignasi35:
Thanks for the detailed explanation!

ignasi35:
@markusjura why is `http://:9000/user` even published in the service gateway if that's a private API? I would expect ` user -> [<list of (ip/port)] ` to exist in a Service Registry that's only consumable from the private network.

[...]

PR #78, opened 2017-04-13
[WIP] Added support for Maven

jroper:
At this stage, devmode works.

* Created pom files for all projects.
* Disabled the PlayLayoutPlugin for the web-gateway, so that it would be a standard layout that Maven could easily build, and restructured the project accordingly.
* Modified the web-gateway to not depend on Play's routes compiler, currently there's no support for compiling Play routes files in Maven.  This meant doing the following:
    - Switching to using the Java RoutingDsl.
    - Extracting query parameters from the request directly, since the Java RoutingDsl doesn't support automatically extracting them.
    - Using hard coded paths, rather than the reverse router, since that no longer exists.
* Modified the web-gateway to not depend on sbt-web support. The only place we were really using sbt web was for extracting web-jar assets so they were available under the lib path. This has been replaced with using the play-webjars support, which locates webjar assets directly from the webjars themselves.


user-impl/src/bundle/bundle.conf
```diff
@@ -0,0 +1,28 @@
...
+components = {
+  user = {
+    description = "user"
+    file-system-type = "universal"
+    start-command = ["user/bin/user"]
+    endpoints = {
+      "user" = {
+        bind-protocol = "http"
+        bind-port = 0
+        services = ["http://:9000/user","http://:9000/api/user?preservePath"]
```
ignasi35: I think this service URI causes an issue in ConductR. ConductR creates a mapping using the service name `user` which, in `online-auction-java` clashes with `web-gateway`'s `/user`. 

The change in `UserService` hasn't been pushed yet since it's still unclear wether ConductR should stop doing that. See https://github.com/lagom/online-auction-java/issues/32

ignasi35: I think `"http://:9000/api/user?preservePath"` should not be listed since `UserSevice` has ACLs disabled.

ignasi35: After reviewing all file changes in the PR I'm not sure what's so special about `user-service-impl` that it's the only project with a hardcoded `bundle.conf`. 
I'm not familiar with ConductR's support for maven but if it doesn't exist then we need `bundle.conf` in all projects, and if it does, there's probably something special un `user-service-impl`.

web-gateway/src/main/twirl/views/createItem.scala.html
```diff
@@ -8,7 +8,7 @@
 
     <h2>@message("createItem")</h2>
 
-    @foundationForm(itemForm, routes.ItemController.createItem()) {
+    @foundationForm(itemForm, play.api.mvc.Call("POST", "/item")) {
```
ignasi35: meh

ignasi35: I mean, I understand the change but the final code due to maven-play limitations is a step back. :-(

ignasi35:
> At this stage, devmode works
🎉

jroper:
I haven't started testing with ConductR - the only reason I added the user services ConductR configuration is that I needed to mvn install it at some point for some reason, which would fail if I didn't have those files there, so it was either that or delete the assembly configuration out of the pom.

pom.xml
```diff
@@ -0,0 +1,148 @@
+<!-- This project has build files for both Maven and sbt.
+  sbt is the recommended build tool for using with Lagom,
+  but Maven can also be used. -->
```
TimMoore: I was under the impression that we wanted to give Maven and sbt equal status.

user-impl/src/assembly/conductr-bundle.xml
```diff
@@ -0,0 +1,32 @@
...
+            </excludes>
+        </fileSet>
+        <fileSet>
+            <outputDirectory>user/bin</outputDirectory>
+            <directory>src/bundle/user/bin</directory>
+            <fileMode>0755</fileMode>
+        </fileSet>
+    </fileSets>
+
+</assembly>
```
TimMoore: Don't all of the other impl projects and web-gateway need this as well?

pom.xml
```diff
@@ -0,0 +1,148 @@
...
+                <artifactId>foundation-icon-fonts</artifactId>
+                <version>d596a3cfb3</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <lagom.version>1.3.2-SNAPSHOT</lagom.version>
```
TimMoore: This can be bumped to 1.3.2 final now, and hopefully that will fix the build.

user-impl/src/assembly/conductr-bundle.xml
jroper: Yep, haven't got up to ConductR support yet.

pom.xml
jroper: I would still recommend sbt over Maven.

ignasi35:
Fixes #21 , correct?

TimMoore:
See also https://github.com/lagom/activator-lagom-java-chirper/pull/52

TimMoore:
It sounds like a native ConductR Maven plugin may be coming soon enough to make it worth waiting for that rather than adding all of the boilerplate manually.

There's still the problem of a Kafka bundle for ConductR (or some other way of running Kafka) to tackle.

user-impl/src/bundle/bundle.conf
TimMoore: There's a path conflict with the search service as well

ignasi35: There are two problems with the search service actually: path conflict and search-api enabling its ACL.

PR #79, opened 2017-04-13
Update to Inline Instructions in web UI

mehulsr:
Added instructions in README.md and changed name of flag in application.conf


ignasi35:
Thanks @mehulsr !

PR #81, opened 2017-04-27
Add instructions for running in ConductR

TimMoore:
Also includes fixes for a few minor ConductR compatibility issues.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,6 +39,7 @@ lazy val itemApi = (project in file("item-api"))
 
 lazy val itemImpl = (project in file("item-impl"))
   .settings(commonSettings: _*)
+  .settings(kafkaSettings: _*)
   .enablePlugins(LagomJava)
   .settings(
     version := "1.0-SNAPSHOT",
@@ -65,6 +66,7 @@ lazy val biddingApi = (project in file("bidding-api"))
 
 lazy val biddingImpl = (project in file("bidding-impl"))
   .settings(commonSettings: _*)
+  .settings(kafkaSettings: _*)
   .enablePlugins(LagomJava)
   .dependsOn(biddingApi, itemApi)
   .settings(
@@ -91,6 +93,7 @@ lazy val searchApi = (project in file("search-api"))
 
 lazy val searchImpl = (project in file("search-impl"))
   .settings(commonSettings: _*)
+  .settings(kafkaSettings: _*)
   .enablePlugins(LagomJava)
   .settings(
     version := "1.0-SNAPSHOT",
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -191,7 +194,16 @@ def elasticsearch : String = {
...
+  javacOptions in Compile ++= Seq("-encoding", "UTF-8", "-source", "1.8"),
+  javacOptions in (Compile, compile) ++= Seq("-Xlint:unchecked", "-Xlint:deprecation", "-parameters")
+)
+
+// Include this into impl projects that use the message broker API
+// It overrides the production configuration to use a hardcoded Kafka broker
+// host and port rather than looking it up from the service locator.
+// See docs/running-in-conductr.md for details.
+def kafkaSettings: Seq[Setting[_]] = Seq(
+  BundleKeys.startCommand += "-Dlagom.broker.kafka.service-name=''"
```
ignasi35: 👍

ignasi35: Don't we need a broker list too ?

TimMoore: It happens to default to localhost:9092: https://github.com/lagom/lagom/blob/master/service/core/kafka/client/src/main/resources/reference.conf#L4

ignasi35: How about making the broker explicit here with a note:

```
// You may have to edit this list if your Kafka Docker container is 
// not listening on 127.0.0.1:9092
```

I'd rather add this info here than in the MD file with instructions you introduced in this PR to keep that file at a minimum. Also, when the kafka bundle for ConductR is ready it'll be possible to come back here and remove this whole setup without changing other files.

WDYT?

TimMoore: Sure

search-api/src/main/java/com/example/auction/search/api/SearchService.java
```diff
@@ -22,6 +22,6 @@ default Descriptor descriptor() {
                 Service.restCall(Method.GET, "/search?pageNo&pageSize", this::search)
         ).withPathParamSerializer(
                 UUID.class, PathParamSerializers.optional("UUID", UUID::fromString, UUID::toString)
-        ).withAutoAcl(true);
+        );
```
ignasi35: 👏

build.sbt
```diff
@@ -203,7 +203,12 @@ def commonSettings: Seq[Setting[_]] = eclipseSettings ++ Seq(
 // host and port rather than looking it up from the service locator.
 // See docs/running-in-conductr.md for details.
 def kafkaSettings: Seq[Setting[_]] = Seq(
-  BundleKeys.startCommand += "-Dlagom.broker.kafka.service-name=''"
+  BundleKeys.startCommand ++= Seq(
+    "-Dlagom.broker.kafka.service-name=''",
+    // You may have to edit this list if your Kafka
+    // server is not listening on 127.0.0.1:9092
+    "-Dlagom.broker.kafka.brokers='127.0.0.1:9092'"
+  )
```
ignasi35: After merging I realised that I?m not sure this is the recommended approach for production users to override Kafka settings. 
Should we document this as a best practice in the docs for the use case where a developer wants to override the settings, or should we make it obvious here that this is a hack to ease the usage of the `online-auction` sample app?
see https://github.com/lagom/lagom/issues/702

TimMoore: Do you mean you're not sure if setting it in this particular ConductR setting is the recommended approach?

I'm not sure, either. When I asked about it in Slack, I didn't get any responses, and the Play guide offered a number of different options but didn't talk about ConductR specifically.

I based this approach on a little snippet of code I found in the [`sbt-conductr` README](https://github.com/typesafehub/sbt-conductr#adding-start-command-options). I'm not sure if this is best practice.

If it were more than a small number of properties, I would probably want to put them into a separate `.conf` file and pass `-Dconfig.resource` instead, as described in https://www.playframework.com/documentation/2.5.x/ProductionConfiguration#Specifying-an-alternate-configuration-file

I think #702 will be a better way to handle this in the long run.

PR #89, opened 2017-05-24
Fix link in 'Running in Production Mode' section of README.md

longmuir:
Looks like it needed updating.


ignasi35:
Good cathc. Thanks @longmuir

PR #90, opened 2017-05-30
Remove custom optional serializer

### LINKED ISSUE #75: Remove custom optional serializer
#11 introduced a custom serializer for java Optional (see https://github.com/lagom/online-auction-java/blob/master/item-impl/src/main/resources/application.conf#L11-L19 and related class/es).

When this is provided by Akka the custom serializer should be removed and the provided by used instead.

lakhina:
fix #75 by using `akka.remote.serialization.MiscMessageSerializer` instead of custom optional serializer.


lakhina:
@ignasi35 Can you please confirm I am moving right way?

ignasi35:
HI @lakhina I don't think we need any serializer. If this app is using the latest lagom, the optional serializer should be picked up, doesn't it?
I haven't looked into it in a while so I trust your tests. Does it fail without `MiscMessageSerializer ` ?

lakhina:
@ignasi35 I haven't tested without `MiscMessageSerializer` but i took reference from https://github.com/akka/akka/blob/master/akka-remote/src/main/resources/reference.conf#L51

lakhina:
I felt if akka is using it then there should be a need for it.

lakhina:
@ignasi35  Thanks. It is working good without `MiscMessageSerializer`

ignasi35:
Good job!

PR #91, opened 2017-05-30
[WIP] Set up Transaction Service

yg-apaza:
Use Case 2: As an auction winner I want to init the transaction after the auction finished


jroper:
This looks like a good start, I merged because I don't see any reason not to.

PR #92, opened 2017-05-31
Login implementation initial phase

lakhina:
-  Add seperate route for login form and request
-  Add Auth model
-  UserCommand instance variable update
-  Implement user authentication call


lakhina:
@ignasi35 @TimMoore @jroper  Please review the initial implementation of Login Service

user-impl/src/main/java/com/example/auction/user/impl/UserEntity.java
```diff
@@ -44,7 +44,7 @@ private Behavior notCreated() {
         );
 
         b.setCommandHandler(CreateUser.class, (create, ctx) -> {
-            User user = new User(UUID.fromString(entityId()), create.getName());
+            User user = new User(UUID.fromString(entityId()), create.getName(),"admin","admin");
```
jroper: What I would do here is have the username as a property of user, but have the password as its own event, so you'd persist two events here, a `UserCreated` event, followed by a `PasswordUpdated` event.

In addition the `PasswordUpdated` event should store the password as a hash.  Bcrypt is good for hashing passwords:

http://www.mindrot.org/projects/jBCrypt/

lakhina: @jroper I will persist this as two seperate events `UserCreated` and `PasswordUpdated` in my new commits.

lakhina: @jroper Do you to store password in a separate table or in the same table you want to store the encrypted password?

ignasi35:
Hi @lakhina did you close this by mistake? Are you moving your work to #95  ?

lakhina:
@ignasi35 I am moving work to #95 as I had to use new branch. Reimplementing the whole work with new approach,

PR #94, opened 2017-06-01
Transaction Service - Submit delivery details

yg-apaza:
Use Case 3: As a buyer, I want to submit delivery details for the seller

- [x] Update `TransactionEntity`, command: `SubmitDeliveryDetails`, event: `DeliveryDetailsSubmitted` and state
- [x] Add tests: `shouldForbidSubmittingDeliveryDetailsByNonBuyer`, `shouldEmitEventWhenSubmittingDeliveryDetails`
- [x] UI and Controllers on `web-gateway`
- [x] Add `getTransaction` and fill form of DeliveryDetails in UI
- [x] Show transaction details
- [x] Add integration tests (`shouldCreateTransactionOnAuctionFinished`, `shouldNotCreateTransactionWithNoWinner`, `shouldSubmitDeliveryDetails`)


yg-apaza:
@jroper @TimMoore

- I got failed test from [latest build on Travis CI](https://travis-ci.org/lagom/online-auction-java/builds/238611274#L3647), why is this?
- Should I use [`commandFailed`](https://github.com/lagom/online-auction-java/pull/94/files#diff-5d960945968560c73aae8be2f3ca2bf3R65) method or instead `throw new Forbidden(...)`?
- I was looking for some examples in other services and found 2 TODOs: [Use Forbidden instance](https://github.com/lagom/online-auction-java/blob/master/item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java#L75) and [Avoid using TransportException
](https://github.com/lagom/online-auction-java/blob/master/item-impl/src/main/java/com/example/auction/item/impl/PItemEntity.java#L50)Aren't these contradictory?

yg-apaza:
@TimMoore @jroper This task is finished, please review

yg-apaza:
This PR is now finished :sweat_smile: 
@jroper @TimMoore @ignasi35 , could you please review?

ignasi35:
On it...

transaction-impl/src/main/java/com/example/auction/transaction/impl/DeliveryData.java
```diff
@@ -0,0 +1,24 @@
+package com.example.auction.transaction.impl;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.Value;
+
+@Value
+public class DeliveryData {
```
ignasi35: This is the private representation of the `transaction-api`'s `DeliveryInfo`.

We have a terrible mix of naming convetions in that sometimes the private copy is `PXyz`, sometimes it's the exact same name with only a package difference, ...
In general I hate using `Data` and `Info` for anything other than signifying that `data` is master and `info` can be derived from `data`.

I don't think this issue of naming should be solved in this PR but it' be good if we settled and considered creating a convention.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
```diff
@@ -28,25 +33,66 @@ public Behavior initialBehavior(Optional<TransactionState> snapshotState) {
...
+        );
+
+        builder.setCommandHandler(SubmitDeliveryDetails.class, (cmd, ctx) -> {
+            if(cmd.getUserId().equals(state().getTransaction().get().getWinner())) {
+                return ctx.thenPersist(new DeliveryDetailsSubmitted(entityUUID(), cmd.getDeliveryData()), (e) ->
+                        ctx.reply(Done.getInstance())
+                );
+            }
+            else
+                throw new Forbidden("Only the auction winner can submit delivery details");
```
ignasi35: I think you'd better use `ctx.failed` here. Throwing an exception will kill the underlying actor and cause it to be rebuilt. If you use  `ctx.failed` you are signaling the command process completed successfully with a rejection.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
```diff
@@ -0,0 +1,49 @@
...
+                    data.get().getAddressLine2(),
+                    data.get().getCity(),
+                    data.get().getState(),
+                    data.get().getPostalCode(),
+                    data.get().getCountry()
+            ));
+        }
+        else
+            return Optional.empty();
+    }
```
ignasi35: This is a very common antipattern of `Optional`. A `if(opt.isPresent) {...} else {Optional.empty}` is equivalent to `Optional.map()`. The code can be rewritten as:

```java
        return data.map(deliveryData -> new DeliveryInfo(
                deliveryData.getAddressLine1(),
                deliveryData.getAddressLine2(),
                deliveryData.getCity(),
                deliveryData.getState(),
                deliveryData.getPostalCode(),
                deliveryData.getCountry()
        ));
```

NOTE: IntelliJ made the suggestion and the rewrite for me. IntelliJ detected the antipattern and marked the code in brown, than using `Alt-return` made the fix.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
```diff
@@ -0,0 +1,49 @@
...
+                data.getAddressLine1(),
+                data.getAddressLine2(),
+                data.getCity(),
+                data.getState(),
+                data.getPostalCode(),
+                data.getCountry()
+        );
+    }
+
+    public static TransactionInfo toApi(TransactionState data) {
```
ignasi35: I think this is dead code. Is that correct? Is that WIP?

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
```diff
@@ -0,0 +1,49 @@
...
+        Transaction transaction = data.getTransaction().get();
```
ignasi35: `get()` could throw a `NullPointerException`.
Should this code consider the case where `data.getTransaction()` returns `empty`? If that can't happen, then It'd be great to have a comment: `// this code is always executed when transactions is set so we can get() safely`.

ignasi35:
This is great work and you are progressing steadily!
I haven't reviewed the UI/Play portion but it looks awesome.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
ignasi35: Changing this will break your tests.
Also, I've just noticed there's a similar case few lines below this.

ignasi35:
> I was looking for some examples in other services and found 2 TODOs: Use Forbidden instance and Avoid using TransportException. Aren't these contradictory?

They are.

So far the approach we're using is: if the service is mall enough then use some transport exceptions inside the Persistent Entity. If you think using more domain-rich exceptions will make the code more maintainable then don't.

ignasi35:
Hi @yg-apaza earlier I forgot to mention that this PR doesn't have any integration test (start a server, send some requests and/or messages via a topic stub). Is that on purpose?
Would it be benefitial to add a couple of happy-path tests as we move forward with this service ?

transaction-impl/src/main/java/com/example/auction/transaction/impl/DeliveryData.java
yg-apaza: Raised #98

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
yg-apaza: I forgot to delete it :sweat_smile:

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
jroper: I'm not sure if that's the case - I think all commands are processed in a `try catch`, and when a command handler throws an exception, Lagom catches it and passes it to `ctx.failed` for you.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
yg-apaza: Nevermind, I'm using `TransactionInfo toApi(TransactionState data)`[here](https://github.com/lagom/online-auction-java/pull/94/files#diff-174db08e7c4c614d7a77741a444e5167R69)

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
yg-apaza: So, can I throw an exception? Using `ctx.failed` breaks test, is it related with lagom/lagom#325?

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionMappers.java
yg-apaza: I added a comment, using `get()` is safe

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
yg-apaza: While browsing the Lagom source code, found [this](https://github.com/lagom/lagom/blob/master/persistence/javadsl/src/main/scala/com/lightbend/lagom/internal/javadsl/persistence/PersistentEntityActor.scala#L178)
As @jroper said, commands are inside a `try catch`, so I think it's safe throwing an exception
I changed this in the latest commit

yg-apaza:
How about now? @ignasi35 @TimMoore @jroper

transaction-impl/src/test/java/com/example/auction/transaction/impl/TransactionServiceImplIntegrationTest.java
```diff
@@ -81,26 +83,47 @@ public void shouldCreateTransactionOnAuctionFinished() {
...
+                    .get(5, SECONDS);
 
-        eventually(new FiniteDuration(10, SECONDS), () -> {
             TransactionInfo retrievedTransaction = transactionService.getTransaction(itemId)
                     .handleRequestHeader(authenticate(creatorId))
                     .invoke()
                     .toCompletableFuture()
                     .get(5, SECONDS);
+
             assertEquals(retrievedTransaction, transactionInfoWithDelivery);
```
ignasi35: I'm not sure this test is correct.
I mean, the test does pass, but it uses `eventually` in a way that's not safe and which we shouldn't promote:
 - eventually is a loop that is executing a block of code with an assertion that _eventually_ becomes successful. The key thing is that the code block is run over and over, which means the code in `transactionService.submitDeliveryDetails...` is invoked several times. IIUC, that code is not idempotent in the `TransactionServiceImpl + TransactionEntity` classes.

ignasi35:
Other than my previous comment it LGTM

transaction-impl/src/test/java/com/example/auction/transaction/impl/TransactionServiceImplIntegrationTest.java
```diff
@@ -108,16 +108,16 @@ public void shouldNotCreateTransactionWithNoWinner() throws Throwable {
...
-                    .invoke(deliveryInfo)
-                    .toCompletableFuture()
-                    .get(5, SECONDS);
+        transactionService.submitDeliveryDetails(itemId)
+                .handleRequestHeader(authenticate(winnerId))
+                .invoke(deliveryInfo)
+                .toCompletableFuture()
+                .get(5, SECONDS);
 
+        eventually(new FiniteDuration(15, SECONDS), () -> {
```
ignasi35: 👍

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
TimMoore: I think `ctx.failed` would be a better example of what we consider best practices... I'm not sure why it's failing the test, but we can look at it in a separate pull request. This is OK for now.

web-gateway/app/controllers/TransactionController.java
```diff
@@ -0,0 +1,147 @@
...
+                                        !transaction.getCreator().equals(user),
+                                        itemId,
+                                        formFactory.form(DeliveryDetailsForm.class).fill(form),
+                                        transaction.getStatus(),
+                                        Optional.empty(),
+                                        nav)
+                        );
+                    }
+                    else {
+                        String msg = ((TransportException) exception.getCause()).exceptionMessage().detail();
```
TimMoore: I'm not sure that it's safe to always assume that `exception.getCause()` will always be a `TransportException`.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
yg-apaza: Since some services use `throw new(...)` and others `ctx.commandFailed`, I raised #108

web-gateway/app/controllers/TransactionController.java
yg-apaza: I'm going to fix this as part of #107

PR #95, opened 2017-06-06
[WIP] Login Implementation 

lakhina:
- Switched the previous login implementation to that using auth event
- Add username and password fields to create user form


lakhina:
I checked with Dbeaver. Auth events are getting saved but problem is I am not able to get them back. Can you please enlighten me, where I am going wrong?

lakhina:
@ignasi35 Can you please help?

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -45,14 +45,34 @@ public UserServiceImpl(PersistentEntityRegistry registry, ActorSystem system, Ma
     }
 
     @Override
+    public ServiceCall<Auth, Auth> updateAuth() {
+        return auth -> {
+            return authRef(auth.getUsername()).ask(new AuthCommand.UpdateAuth(auth.getId(), auth.getUsername(), auth.getPassword()));
```
ignasi35: why return the `Auth`?

It's an object that contains a clear-text password and we should limit the password exposure.
I think a `Done` is more than enough.

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -45,14 +45,34 @@ public UserServiceImpl(PersistentEntityRegistry registry, ActorSystem system, Ma
...
                             maybeUser.orElseGet(() -> {
                                 throw new NotFound("User " + userId + " not found");
                             })
-            );
+                    );
+        };
+    }
+
+    @Override
+    public ServiceCall<Auth, Auth> login() {
```
ignasi35: I think the return type here should be a `Token` or a `Boolean`  or a `Done`. See commend above.

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -68,11 +88,18 @@ public UserServiceImpl(PersistentEntityRegistry registry, ActorSystem system, Ma
...
         return entityRef(id.toString());
     }
 
     private PersistentEntityRef<UserCommand> entityRef(String id) {
+        registry.register(UserEntity.class);
         return registry.refFor(UserEntity.class, id);
     }
+
+    private PersistentEntityRef<AuthCommand> authRef(String username) {
+        registry.register(AuthEntity.class);
```
ignasi35: As a general rule, do all `registry.register(XxxYyZzzzz.class);` on the `MyServiceImpl`  constructor and just once. This is registering `AuthEntity.class` over and over. And same thing applies on line 97.

user-impl/src/main/java/com/example/auction/user/impl/AuthEvent.java
```diff
@@ -0,0 +1,24 @@
+package com.example.auction.user.impl;
+import com.example.auction.user.api.Auth;
+import com.example.auction.user.api.User;
+import com.lightbend.lagom.javadsl.persistence.AggregateEvent;
```
ignasi35: As a general rule, avoid reusing API classes in PersistentEntity-related operations. This is a best practice knonw as Anti-corruption layer.

In authentication in particular you will find that almost every message exchanged to-from the service differs. For example, creating credentials may need two passwords, a username and some other data but logging in only needs one password and a username. Then, in no case should the service send a password in a response...
... you get the gist ;-)

user-impl/src/main/java/com/example/auction/user/impl/AuthCommand.java
```diff
@@ -0,0 +1,31 @@
...
+import com.example.auction.user.api.Auth;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.lightbend.lagom.javadsl.persistence.PersistentEntity;
+import com.lightbend.lagom.serialization.Jsonable;
+import lombok.Value;
+
+import java.util.Optional;
+import java.util.UUID;
+
+public interface AuthCommand extends Jsonable {
```
ignasi35: Let's try to avoid `Auth` altogether. It is confusing because it could mean `Authentication` or `Authorization`.

In credential management it is very difficult to get naming correct.
In the past I found that a set of strings that a user sends to a server as an authentication could be named `Credentials`. So I'd have a `UsernamePasswordCredentials` or `Oauth2TokenCredentials` implementations.

Then, the data a user sends in a signup form are not yet credentials because we don't really know if they are valid: the username could be a duplicate, or the password too weak.

Etc...

lakhina:
@ignasi35 @TimMoore @jroper 
Initially it started working fine before adding  Credential in implementation layer but now it is throwing given exception:
`com.lightbend.lagom.javadsl.persistence.PersistentEntity$UnhandledCommandException: Unhandled command [com.example.auction.user.impl.CredentialCommand$UpdateCredential] in [com.example.auction.user.impl.CredentialEntity] with id [asdfg]`

I am not being able to resolve. Please help .

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -45,14 +49,42 @@ public UserServiceImpl(PersistentEntityRegistry registry, ActorSystem system, Ma
...
+        };
+    }
+
+    @Override
+    public ServiceCall<Credential, String> login() {
+        return (Credential req) -> {
+            return credRef(req.getUsername()).ask(CredentialCommand.GetCredential.INSTANCE)
+                    .thenApply(maybeCredential-> {
+                                 if(maybeCredential.isPresent()) {
+                                     if (BCrypt.checkpw(req.getPassword(), maybeCredential.get().getPassword())){
```
jroper: Good.

user-impl/src/main/java/com/example/auction/user/impl/CredentialEntity.java
```diff
@@ -0,0 +1,45 @@
+package com.example.auction.user.impl;
+
+import com.lightbend.lagom.javadsl.persistence.PersistentEntity;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+public class CredentialEntity extends PersistentEntity<CredentialCommand, CredentialEvent, Optional<PCredential>> {
```
jroper: I'm not sure the reason for separating the credential entity out from the user entity - perhaps I wasn't clear enough before. I meant that update credentials should be a new event in the user entity, not that there should be two different entities. I think if you copy `UpdateCredential` into `UserCommand`, and `CredentialUpdated` into `UserEvent`, and then the behaviour in this class can be copied into the `UserEntity`, that will achieve what I was thinking.

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -45,14 +49,42 @@ public UserServiceImpl(PersistentEntityRegistry registry, ActorSystem system, Ma
...
                             })
-            );
+                    );
+        };
+    }
+
+    @Override
+    public ServiceCall<Credential, String> login() {
+        return (Credential req) -> {
+            return credRef(req.getUsername()).ask(CredentialCommand.GetCredential.INSTANCE)
```
jroper: Perhaps rather than doing this logic here, instead create a `Login` command instead of a `GetCredential` command. Then, the login business logic can be done in the persistent entity.

Then again, someone may disagree with me, @ignasi35 or @TimMoore what do you think?

user-impl/src/main/java/com/example/auction/user/impl/CredentialEntity.java
ignasi35: I agree with having the logic in UserEntity. 

In past comments I simply stated `Credentials` is sometimes a more adequate name and that probably misslead @lakhina . Apologies for the confusion.

lakhina: @jroper @ignasi35 I seprated those events because. user registry save and retrieve on basis og userid but credential registry can not work on basis of that. We need to match password corresponding to username and find user id corresponding to that.

lakhina: @jroper @ignasi35 please let me know if my thought isn't right

build.sbt
```diff
@@ -162,7 +162,10 @@ lazy val userImpl = (project in file("user-impl"))
   .dependsOn(userApi)
   .settings(
     version := "1.0-SNAPSHOT",
-    libraryDependencies += lagomJavadslPersistenceCassandra
+    libraryDependencies ++= Seq(
+      lagomJavadslPersistenceCassandra,
+      "org.mindrot" % "jbcrypt" % "0.3m"
```
TimMoore: It looks like there is a more recent version 0.4. It is probably a good idea to use that instead.s

user-impl/src/main/java/com/example/auction/user/impl/CredentialCommand.java
```diff
@@ -0,0 +1,50 @@
...
+
+        @JsonCreator
+        public UpdateCredential(UUID id, String username, String password) {
+            this.id = id;
+            this.username = username;
+            this.password = password;
+        }
+    }
+        public static String hashPassword(String password_plaintext) {
+        String salt = BCrypt.gensalt(workload);
```
TimMoore: Is there a specific reason not to use the default number of hashing rounds?

user-impl/src/main/java/com/example/auction/user/impl/CredentialEntity.java
TimMoore: I think this is a situation where you would use a read-side processor. The credentials can be stored in the user entity, and a read-side processor can update a separate credential registry table with username and password fields. Then the login service call would query that table.

I like to create a class that encapsulates both the update and the query of the read-side table. In this case, you could create a class called `CredentialsRegistry` that registers an event processor in its constructor. This is the same pattern used in [`ItemRepository.java`](https://github.com/lagom/online-auction-java/blob/master/item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java).

web-gateway/app/controllers/CreateUserForm.java
```diff
@@ -14,4 +26,12 @@ public String getName() {
...
         this.name = name;
     }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public void setPassword(String password) {
+        this.password = password;
+    }
```
TimMoore: Most user registration forms include duplicate password and "Verify Password" fields that have to match, to make sure they don't enter the password with a typo and then are unable to log in.

web-gateway/app/controllers/Main.java
```diff
@@ -55,6 +56,7 @@ public Main(Configuration config, MessagesApi messagesApi, UserService userServi
                     }
 
                     return userService.createUser().invoke(new User(form.get().getName())).thenApply(user -> {
+                        userService.updateCredential().invoke(new Credential(user.getId(), form.get().getUsername(), form.get().getPassword()));
```
TimMoore: Consider that there could be a failure in between when `createUser` is invoked and when `updateCredential` is invoked. For example, the server running this code could crash, or the network between the Play app and the user service could go down in between calls. That would leave the user in a half-created state: they wouldn't be able to log in.

TimMoore:
@lakhina thanks for submitting the WIP pull request. My major concern is that this has bitten off too big a problem to solve in one step.

Here's a way you could break it down into a series of smaller changes:

- [ ] Update the "Create User" form to include username, password and verify password fields (don't do anything with them yet)
- [ ] Show an error if the password and verify password fields don't match
- [ ] Add username and password to the `createUser` service call, where the password is hashed, and pass the username and hashed password in the `CreateUser` command and `UserCreated` event
- [ ] Create the `CredentialsRegistry` class and event processor, which stores rows in a table with the user ID, username and password hash
- [ ] Change the `createUser` service call to check the `CredentialsRegistry` to make sure there is no existing user with the same username before creating a new user. *
- [ ] Implement the Login form in the web gateway and the `login` service call in the user service.
- [ ] Add the Log Out button and controller to the web gateway. It isn't necessary to create a logout service call at this time, since it doesn't make any changes to the persistent data, it only clears the Play session.

Each of these steps should also include appropriate tests when it makes sense.

One other thing I would suggest: let's have it use email and password rather than a username. This will let us extend this in the future to include a step where we verify the email address before the account is activated.

\* Note that, because read-side processors in a CQRS system are eventually consistent, it is still possible to end up with duplicates, if two users register with the same username at around the same time. Greg Young, who helped to popularize CQRS architectures, [wrote about this](http://codebetter.com/gregyoung/2010/08/12/eventual-consistency-and-set-validation/). It's a hard problem to solve completely, but there are some "good enough" solutions that we should demonstrate in this app, since this is a very common problem that many of our users ask about. We can talk about the details of that once the basics outlined above are working.

lakhina:
@TimMoore So I should close this PR and open corresponding small PRs?

TimMoore:
Yes, I think doing it one step at a time in separate PRs is better. I'll work on filing issues for each of the steps.

PR #100, opened 2017-06-20
[WIP] Prepares online-auction to Lagom 1.4.0 bump

ignasi35:
This currently requires a local compilation of the https://github.com/lagom/lagom/tree/play-2.6-akka-2.5 branch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,7 +99,6 @@ lazy val searchImpl = (project in file("search-impl"))
   .settings(
     version := "1.0-SNAPSHOT",
     libraryDependencies ++= Seq(
-      lagomJavadslPersistenceCassandra,
       lagomJavadslTestKit,
       lagomJavadslKafkaClient,
       lombok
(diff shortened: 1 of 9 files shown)


ignasi35:
This currently has some issue on the play app: `webGateway`

ignasi35:
> This currently has some issue on the play app: webGateway

The issues are now solved (see 85dce88) but there's still work to be done:

- [ ] use new play default filters (play25 had XSS, CSRF,... disabled by default and play26 adds them by default). I'd be good to have the filters enabled
- [ ] test this over an environment where there's data created with an `online-auction-java` running on Lagom 13x. That would help detect issues for users upgrading a production environment.

TimMoore:
@ignasi35 is it possible to rework this in a way that it is compatible with _both_ Lagom 1.3 and 1.4?

That way, we could merge everything but the version bump, rather than leaving this open as a long-lived branch.

We can then fix new deprecations in a subsequent PR, once Lagom 1.4 is released.

ignasi35:
> @ignasi35 is it possible to rework this in a way that it is compatible with both Lagom 1.3 and 1.4?

It will need some thought. I'm not 100% familiar with all the changes and I think at least the form validation is not compatible.

We can consider this branch as a playground to smoke test and prevent further surprises.

renatocaval:
done by #164

renatocaval:
Done by #164

PR #101, opened 2017-06-21
Add UserRegisteration class to seperate registration details

lakhina:
- Add `email` to `User` of api
- Add `UserRegisteration` to seperate registration details

### DIFF
--- a/user-api/src/main/java/com/example/auction/user/api/User.java
+++ b/user-api/src/main/java/com/example/auction/user/api/User.java
@@ -12,23 +12,27 @@ public final class User {
 
     private final UUID id;
     private final String name;
+    private final String email;
 
     @JsonCreator
-    private User(@JsonProperty("id") Optional<UUID> id, @JsonProperty("name") String name) {
+    private User(@JsonProperty("id") Optional<UUID> id, @JsonProperty("name") String name, @JsonProperty("email") String email) {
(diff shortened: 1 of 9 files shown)


user-api/src/main/java/com/example/auction/user/api/User.java
```diff
@@ -12,23 +12,13 @@
...
+    private final String email;
 
     @JsonCreator
-    private User(@JsonProperty("id") Optional<UUID> id, @JsonProperty("name") String name) {
-        this.id = id.orElse(null);
-        this.name = name;
-    }
-
-    public User(UUID id, String name) {
+    public User(@JsonProperty("id") UUID id, @JsonProperty("name") String name, @JsonProperty("email") String email) {
```
TimMoore: Did you find it necessary to add the `@JsonProperty` annotations, or are you doing it just to be explicit?

We have enabled the Jackson parameter names extension, which can detect the names of constructor parameters compiled into the bytecode, so it shouldn't be necessary to add explicit annotations. If that isn't working, it might be that something is misconfigured.

TimMoore:
As a next step, before going further with the implementation it would be a good idea to add some basic tests for `UserEntity` and `UserServiceImpl`... I see that none were written when `user-impl` was originally created.

You can look at the other service implementation projects for examples to guide you.

PR #102, opened 2017-06-21
Move user related calls to UserController

lakhina:


web-gateway/app/controllers/Main.java
```diff
@@ -38,36 +38,4 @@ public Main(Configuration config, MessagesApi messagesApi, UserService userServi
         );
     }
```
TimMoore: The `userService` field is no longer used, and can be removed.

TimMoore: Same with `formFactory` and `showInlineInstruction`.

web-gateway/app/controllers/UserController.java
```diff
@@ -0,0 +1,79 @@
...
+
+import javax.inject.Inject;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+import static com.example.auction.security.ClientSecurity.authenticate;
```
TimMoore: A lot of these `import` statements aren't actually used in the code. Recommend using an IDE function to automatically organize your imports:

- https://www.jetbrains.com/help/idea/optimizing-imports.html
- https://help.eclipse.org/neon/topic/org.eclipse.jdt.doc.user/gettingStarted/qs-OrganizeImports.htm?cp=1_1_0_3_6

You may need to customize the settings for import order to make sure that the imports don't flip-flop back and forth when different people use different IDEs to edit the file.

Here are the settings in my IDE (the defaults for IntelliJ IDEA):

screen shot 2017-06-22 at 12 01 39 pm

web-gateway/app/controllers/Main.java
lakhina: @TimMoore I removed formFactory and showInlineInstruction but userService need to be there even for index page request.

TimMoore: It needs to be injected into the constructor and passed to `super`, but I don't see anywhere the instance variable is read.

TimMoore:
@lakhina please let us know what your plans are with this. Can you do the clean up I suggested in my comments? IDEA should be able to detect and auto-fix most of these problems. Let me know if you need more specific advice.

lakhina:
@TimMoore I have just optimised the imports in UserController.java Please review and let me know if any changes required

web-gateway/app/controllers/Main.java
TimMoore: What about the suggestion to remove the unused fields from `Main`? IDEA should highlight these, too. F2 will jump between problems that IDEA notices. Alt-Enter will provide options to fix them.

lakhina: @TimMoore  userService can't be removed as it is needed in the index page.

TimMoore: Sorry... I had missed the changes to this file in your most recent commit. I thought only `UserController.java` had changed. Your most recent version is exactly what I had in mind, thanks!

PR #106, opened 2017-06-23
Add read-side processor in transaction service

yg-apaza:


transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
```diff
@@ -0,0 +1,195 @@
...
+                                "ORDER BY itemId DESC " +
+                                "LIMIT ?",
+                        userId,
+                        userId,
+                        status,
+                        limit
+                )
+                .thenApply(List::stream)
+                .thenApply(rows -> rows.skip(offset))
+                .thenApply(rows -> rows.map(TransactionRepository::toTransactionSummary))
```
ignasi35: Hmm, this is using publicAPI. It's fine now but we should keep an eye on this (using same model for public API and DB storage).

yg-apaza: Should I use here something like `PTransactionSummary` and then convert it to `TransactionSummary`?

TimMoore: This isn't used on the write side... this is the code that maps the internal representation to the public API, so I think it's OK. This is just like what's already there in `ItemRepository`. Does that alleviate your concern, @ignasi35?

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
```diff
@@ -0,0 +1,195 @@
...
+                .thenApply(row -> (int) row.get().getLong("count"));
+    }
+
+    private CompletionStage<PSequence<TransactionSummary>> selectTransactionSummary(
+            UUID userId, TransactionInfoStatus status, long offset, int limit) {
+        return session
+                .selectAll(
+                        "SELECT * FROM transactionSummary " +
+                                "WHERE (creatorId = ? OR winnerId = ?) AND status = ? " +
+                                "ORDER BY itemId DESC " +
```
TimMoore: It might be more useful to capture and order by the date the auction was completed. Thinking from the user's perspective, when viewing historical transactions I would expect to see the most recently completed ones first.

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionSummary.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.Value;
+
+import java.util.UUID;
+
+@Value
+public class TransactionSummary {
+
+    UUID itemId;
```
lakhina: @yg-apaza I think it is always better to put access modifiers

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
TimMoore: I just had a look through, and it looks like it will be a lot of extra work to add the auction completion date, so maybe we can do that as a later improvement. For now, ordering by `itemId DESC` is OK, and it's likely to be approximately the same as ordering by the auction completion date. The `itemId` is a time-based UUID, so it will increase over time.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
```diff
@@ -0,0 +1,195 @@
...
+                    "CREATE TABLE IF NOT EXISTS transactionSummary (" +
+                            "itemId timeuuid PRIMARY KEY, " +
+                            "creatorId UUID, " +
+                            "winnerId UUID, " +
+                            "itemTitle text, " +
+                            "currencyId text, " +
+                            "itemPrice int, " +
+                            "status text" +
+                            ") " +
+                            "WITH CLUSTERING ORDER BY (itemId DESC)"
```
TimMoore: As I mentioned on the call, Cassandra is much more strict than most databases about what fields you can use in the `WHERE` clause of a query.  In particular, the primary key needs to match the key you are using to filter your query. So in this case, we do not want the `itemId` to be the primary key, since we are not querying by `itemId`.

Cassandra does not even support `OR` in queries, so querying by `creatorId = ? OR winnerId = ?` unfortunately is not allowed.

Instead, what you can do is create a table called `userTransactions` with a `userId` column as the ["partition key"](http://docs.datastax.com/en/glossary/doc/glossary/gloss_partition_key.html).

Something like this:

```cql
CREATE TABLE IF NOT EXISTS userTransactions (
  userId UUID,
  itemId timeuuid,
  creatorId UUID,
  winnerId UUID,
  itemTitle text,
  currencyId text,
  itemPrice int,
  status text,
  PRIMARY KEY (userId, status, itemId)
)
WITH CLUSTERING ORDER BY (status ASC, itemId DESC)
```

This sets `PRIMARY KEY (userId, status, itemId)` because we need to filter by `userId` and `status`, and order by `status` and `itemId` (`ORDER BY status` is required due to https://issues.apache.org/jira/browse/CASSANDRA-10271)

When processing `TransactionStarted` events, you'll need to insert twice: once with `userId` set to `creatorId` and once with `userId` set to `winnerId`. This lets you avoid the unsupported `OR` clause in the query.

So, creating and querying tables in Cassandra is more complex than PostgreSQL or MySQL, but for that price you get very good performance and scalability. It takes a while and some trial and error to get used to it though. I'm still not quite used to it myself! :smile:

I would recommend downloading [DataStax DevCenter](https://docs.datastax.com/en/developer/devcenter/doc/devcenter/dcInstallation.html) or [DBeaver Enterprise Edition](http://dbeaver.jkiss.org/download/enterprise/). These are GUI tools that let you explore your Cassandra database. You can test creating tables, inserting data into them, and querying them from one of these tools, without the overhead of having to make changes to the Java code each time you want to try something new.

Also, as I said, adding tests for the event processor can help. The tests will let you feed data to the processor without having to finish implementing transactions in the web UI. You can look at [`ItemEventProcessorTest`](https://github.com/lagom/online-auction-java/blob/master/item-impl/src/test/java/com/example/auction/item/impl/ItemEventProcessorTest.java) (which should be renamed to `ItemRepositoryTest`) as a guide.

If you get stuck again, that's OK. Just push what you have and let me know, and I can help out more.

yg-apaza: Thanks @TimMoore, now I don't know where to put the filter (by `userType`: `creator` or `winner`). I was thinking into 3 options:
- Create a Cassandra query: I tried `SELECT * FROM userTransactions WHERE userId = ? AND creatorId = ? ...` to get all transactions of a user where the user is the creator, after executing that query, Cassandra asks for `ALLOW FILTERING` because `creatorId` isn't a partition or cluster key.
- Get all transactions (whether the user is creator or winner) and then manually filter in `TransactionRepository.getTransactionsForUser`. It might be weird mix this with `PaginatedSequence` since we don't know how many transactions has a user as a creator or as a winner, then each page may have different size
- Show all the transactions in `web-gateway`, and hide transactions depending on the filter.

yg-apaza:
Regarding the tests, I'll wait until #121 is reviewed and merged.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
TimMoore: > Show all the transactions in `web-gateway`, and hide transactions in the UI depending on the filter.

@yg-apaza, I had in mind that the UI would show all transactions that I have been a part of, either as the seller or as the winner. To be honest, I hadn't even thought about filtering them, but just displaying them all together, maybe with different formatting for sold items vs. bought items. We haven't really defined very clear requirements for this... it could go either way. :smile:

Maybe for simplicity's sake, let's display all of the transactions in a single unfiltered list for now, since that makes things easier. We might choose to add filtering, or split the page in two, as a future enhancement.

With Cassandra, each type of query you want to do requires a separate table structure (with the possibility of using ["materialized views"](http://docs.datastax.com/en/cql/3.3/cql/cql_using/useCreateMV.html) that automatically derive one table from the data in another). This is one of the ways it differs from relational databases such as MySQL and PostgreSQL... Cassandra requires ["denormalization"](http://www.datastax.com/dev/blog/basic-rules-of-cassandra-data-modeling), or storing the same data in multiple ways. This is usually considered very bad practice for relational databases, partly because it makes things more complicated and it's easy to have bugs that cause the data to fall out of sync. However with Cassandra, this denormalization is what allows it to be so fast and scalable. Lagom's event sourcing read-side design helps to ensure that nothing falls out of sync.

So, this means that the data design needs to closely reflect the way the data is used in the application. For example, if we wanted to have separate queries to get transactions where I am the seller, and transactions where I am the winner, it would require two different tables with two different primary keys (`transactionsByWinner` and `transactionsBySeller`). The read-side processor would need to insert into both. If we wanted to get all transactions mixed together with an optional `userType` filter, we would need to create a `userType` column in the table, and make it part of the primary key.

build.sbt
```diff
@@ -140,7 +140,8 @@ lazy val transactionImpl = (project in file("transaction-impl"))
     libraryDependencies ++= Seq(
       lagomJavadslPersistenceCassandra,
       lagomJavadslTestKit,
-      lagomJavadslKafkaBroker
+      lagomJavadslKafkaBroker,
+      "com.datastax.cassandra" % "cassandra-driver-extras" % "3.0.0"
```
ignasi35: We can probably extract this into a variable. We're using it in few different places now.

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
```diff
@@ -50,10 +54,15 @@
     default Descriptor descriptor() {
         return named("transaction").withCalls(
                 pathCall("/api/transaction/:id", this::submitDeliveryDetails),
-                pathCall("/api/transaction/:id", this::getTransaction)
-        ).withPathParamSerializer(UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString))
-                .withHeaderFilter(SecurityHeaderFilter.INSTANCE);
-
+                pathCall("/api/transaction/:id", this::getTransaction),
+                pathCall("/api/transaction?userId&status&pageNo&pageSize", this::getTransactionsForUser)
```
ignasi35: is that `userId` meant to be in the queryString?
Hmmm, it looks like the quesryString declares 4 placeholders but using the wrong names.

If this was used by a third party client (non lagom client) it could be confusing for the developer.

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionSummary.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.UUID;
+
+@Value
+public class TransactionSummary {
+
+    private final UUID userId;
+    private final UUID itemId;
+    private final UUID creatorId;
+    private final UUID winnerId;
+    private final String itemTitle;
```
ignasi35: Isn't there one `UUID` too much?
It makes sense to have the request user, the creator and the winner in a command so that you can validate if the command can be fulfilled, but this is a `Summary` where the user making the request won't make much sense, will it?

Put another way: the creator and the winner should see the exact same `TransactionSummary` when requesting it.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
ignasi35: Sorry for the delay. 
I think it's weird that a private method on a non-publicly facing class uses the public API. I'd find it awkward that changing the public API (imagine I renamed `TransactionSummary`) caused a change in `TransactionRepository`.

We can fix this later.

TimMoore:
Looks good. #121 has been merged now so tests can be added 😄

transaction-impl/src/test/java/com/example/auction/transaction/impl/TransactionRepositoryTest.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static com.lightbend.lagom.javadsl.testkit.ServiceTest.bind;
+import static com.lightbend.lagom.javadsl.testkit.ServiceTest.defaultSetup;
+import static org.junit.Assert.assertEquals;
+
+public class TransactionRepositoryTest {
```
TimMoore: This is a great start to the tests, but it does not yet test all of the functionality of `TransactionRepository`.

- There should be a test that calling `getTransactionsForUser` with the `winnerId` works, as well as the `creatorId`.
- There should be a test for pagination, like in `ItemRepositoryTest`

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionRepository.java
```diff
@@ -0,0 +1,209 @@
...
+                            "?, " + // itemPrice
+                            "?" +   // status
+                            ")"
+                    )
+                    .thenApply(accept(s -> insertUserTransactionsStatement = s));
+        }
+
+        private CompletionStage<List<BoundStatement>> insertUserTransactions(UUID itemId, Transaction transaction) {
+            return CassandraReadSide.completedStatements(
+                    Arrays.asList(
```
TimMoore: This could make use of `CassandraReadSideUtils.completedStatements` to avoid the need to call `Arrays.asList`. This is exactly what it's there for! 😄

yg-apaza: Recently, I made a PR updating these changes to core: lagom/lagom#873
How can I get this changes in my local machine?

TimMoore: @yg-apaza it will require that we make a new release with those changes in it. I plan to release Lagom 1.3.6 later this week. After that, we can update Online Auction to use it.

In the meantime, you can use the existing version in `CassandraReadSideUtils`, and then we can update it to use the new method in a future pull request.

yg-apaza:
I think it's done, WDYT?

build.sbt
```diff
@@ -154,6 +154,8 @@ lazy val transactionImpl = (project in file("transaction-impl"))
     transactionApi,
     itemApi,
     tools,
+    // TODO: Remove this dependency when completedStatement with varargs it's included on lagom core
```
TimMoore: Oh I didn't realize that `CassandraReadSideUtils` hadn't been moved to `tools`. Otherwise I would have suggested doing that, too. That's OK, this is only temporary until Lagom 1.3.6 is released (before the end of the week).

PR #112, opened 2017-06-26
 Add User entity test

lakhina:


TimMoore:
I think you can fix the test failure by making `PUser` implement `Jsonable` (like [`PItem`](https://github.com/lagom/online-auction-java/blob/master/item-impl/src/main/java/com/example/auction/item/impl/PItem.java) does).

It's necessary to implement `Jsonable` in `PersistentEntity` command, event and state classes so that Lagom & Akka know to use the Jackson JSON serializer to convert the object to its serialized form for storage in the database or transfer over the network.

lakhina:
@TimMoore Thanks. It worked

lakhina:
@TimMoore Please review

build.sbt
```diff
@@ -162,7 +162,10 @@ lazy val userImpl = (project in file("user-impl"))
   .dependsOn(userApi)
   .settings(
     version := "1.0-SNAPSHOT",
-    libraryDependencies += lagomJavadslPersistenceCassandra
+    libraryDependencies ++= Seq(
+      lagomJavadslPersistenceCassandra,
+      lagomJavadslTestKit
+         )
```
TimMoore: Watch out for indentation... not a big problem of course, but it helps readability to be consistent. IDEA can auto-indent for you: the default shortcut is control-alt-i (control-option-i on Mac).

PR #117, opened 2017-06-29
Add User service implementation test

lakhina:


user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+    private final String password = "admin";
+
+
+    @Test
+    public void shouldBeAbleToCreateUsers() throws Exception {
+        withServer(defaultSetup().withCassandra(true), server -> {
+            UserService userService = server.client(UserService.class);
+            UserRegistration userRegistration= new UserRegistration(name,email,password);
+           userService.createUser().invoke(userRegistration).toCompletableFuture().get(10, SECONDS);
+
```
TimMoore: This is a great start, but this test only tests that `createUser` succeeds, and doesn't test anything about what it actually does.

I would suggest two changes:

1. Check that the returned `User` has the right name and email
2. Check that you can call `getUser` afterwards with the returned user ID, and get back the same data

Also, in the interest of keeping things clean, `id1` is not used, so it can be removed.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+    private final String name = "admin";
+    private final String email = "admin@gmail.com";
+    private final String password = "admin";
+
+
+    @Test
+    public void shouldBeAbleToCreateUsers() throws Exception {
+        withServer(defaultSetup().withCassandra(true), server -> {
+            UserService userService = server.client(UserService.class);
+            User user = new User(id, name, email);
```
TimMoore: `user` is a bit redundant, it is only used to put the `name` and `email` inside it, then take them back out again a few lines later. The `id` is not used at all.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+            UserRegistration userRegistration = new UserRegistration(name, email, password);
+            User user1 = userService.createUser().invoke(userRegistration).toCompletableFuture().get(10, SECONDS);
```
TimMoore: `user1` is not the most descriptive name. A better name would be something like `createdUser`.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+            assertEquals(user.getName(), user1.getName());
+            assertEquals(user.getEmail(), user1.getEmail());
```
TimMoore: If you take the suggestions above, these could be written as:

```java
assertEquals(name, createdUser.getName())
assertEquals(email, createdUser.getEmail())
```

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+            userService.getUser(user1.getId()).invoke().toCompletableFuture().get(10, SECONDS);
```
TimMoore: Again, this only tests that the call succeeds, not its behavior.

You could write something like this to ensure the created user is returned correctly:

```java
assertEquals(createdUser, userService.getUser(createdUser.getId()).invoke().toCompletableFuture().get(10, SECONDS);
```

lakhina:
@TimMoore  I have made the suggested changes. Please review

Issue #115, opened 2017-06-28
Fresh master clone can't be used


ignasi35:
A fresh clone of `master` is not runnable:

1. clone
2. `sbt runAll`
3.  create a user

**Expected** 

Default user home is displayed

**Actual**

```

[warn] o.a.k.c.NetworkClient - Error while fetching metadata with correlation id 1 : {item-ItemEvent=LEADER_NOT_AVAILABLE}
[warn] o.a.k.c.NetworkClient - Error while fetching metadata with correlation id 2 : {item-ItemEvent=LEADER_NOT_AVAILABLE}
[warn] o.a.k.c.NetworkClient - Error while fetching metadata with correlation id 1 : {bidding-BidEvent=LEADER_NOT_AVAILABLE}
[info] o.h.v.i.u.Version - HV000001: Hibernate Validator 5.2.4.Final
[info] Service webGateway listening for HTTP on 0:0:0:0:0:0:0:0:60496
[info] Service searchImpl listening for HTTP on 0:0:0:0:0:0:0:0:55320
[info] Service itemImpl listening for HTTP on 0:0:0:0:0:0:0:0:53203
[info] Service transactionImpl listening for HTTP on 0:0:0:0:0:0:0:0:51742
[info] Service biddingImpl listening for HTTP on 0:0:0:0:0:0:0:0:58427
[info] Service userImpl listening for HTTP on 0:0:0:0:0:0:0:0:56181
[info] (Services started, press enter to stop and go back to the console...)
[error] application - 

! @74f86hh4b - Internal server error, for (GET) [/my/items/created] ->
 
play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[CompletionException: java.util.NoSuchElementException: No value present]]
        at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:293)
        at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:220)
        at play.api.GlobalSettings$class.onError(GlobalSettings.scala:160)
        at play.api.DefaultGlobal$.onError(GlobalSettings.scala:188)
        at play.api.http.GlobalSettingsHttpErrorHandler.onServerError(HttpErrorHandler.scala:100)
        at play.core.server.netty.PlayRequestHandler$$anonfun$2$$anonfun$apply$1.applyOrElse(PlayRequestHandler.scala:100)
[...]

PR #118, opened 2017-06-29
Update getUsers service call in UserServiceImpl

lakhina:


yg-apaza:
I tested on my local machine and @lakhina's changes fix #115 
Problem was due to name change from `UserEntity` to `PUserEntity`
I think it's safe to merge

PR #122, opened 2017-07-03
Upgrade code to make compatible with Lagom 1.4

ignasi35:
This applies changes from https://github.com/lagom/online-auction-java/pull/100 which are compatible with Lagom 1.3 so that final upgrade to Lagom 1.4 is as small as possible.

### DIFF
--- a/web-gateway/app/controllers/ItemController.java
+++ b/web-gateway/app/controllers/ItemController.java
@@ -57,8 +57,8 @@ public class ItemController extends AbstractController {
     public CompletionStage<Result> createItemForm() {
         return requireUser(ctx(), user ->
                 loadNav(user).thenApplyAsync(nav ->
-                                ok(views.html.createItem.render(showInlineInstruction, formFactory.form(ItemForm.class).fill(new ItemForm()), nav))
-                        , ec.current())
+                                ok(views.html.createItem.render(showInlineInstruction, formFactory.form(ItemForm.class).fill(new ItemForm()), nav)),
+                        ec.current())
         );
     }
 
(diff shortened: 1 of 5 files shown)


ignasi35:
I could apply more than I expected: I only left `Validatable` out (validation in Forms).

web-gateway/app/controllers/ItemController.java
```diff
@@ -33,23 +35,30 @@
     private final BiddingService bidService;
 
     private final Boolean showInlineInstruction;
+    private HttpExecutionContext ec;
```
TimMoore: Did you ask anyone on the Play team about the need to specify an explicit `HttpExecutionContext`? It seems like it has created a very ugly API 😦

ignasi35: Yeah, I did. This was added in Play 2.5 but it didn't create any deprecation warning.
There are now separate Execution contexts and you must use the HTTP one for http rendering.

lakhina: Agree on its weird nature

web-gateway/app/controllers/SearchController.java
```diff
@@ -29,34 +31,38 @@
     private final SearchService searchService;
 
     private final Boolean showInlineInstruction;
+    private HttpExecutionContext ec;
 
     @Inject
-    public SearchController(Configuration config,
+    public SearchController(Config config,
```
lakhina: Shouldn't any lib avoid this kind of changes?

lakhina:
I have got to understand what is PR meant to say but not able to understand reason for this complexity

web-gateway/app/controllers/SearchController.java
ignasi35: Yes, and they did.

`Configuration` may still be used but is deprecated. In order to reduced the number of deprecation warnings I made the update to the suggested new API.

TimMoore:
@lakhina these changes were all made for compatibility with Play 2.6 (which Lagom 1.4 depends on).

ignasi35:
There's enough `approve`, I'll merge :-P

PR #126, opened 2017-07-07
Fixes startup issue caused by removal of play.api.Configuration

ignasi35:
When adding changes in `master` in preparation for bump to Lagom 1.4.0 I introduced a runtime error deeming the `web-gateway` unusable.


ignasi35:
This PR adds a `ConfigProvider`  that will publish a `Config` into Guice so that new Controllers can use that instead of `play.api.Configurator` (which is deprecated but available).

yg-apaza:
I tested this and compare with current master branch, now there aren't any errors

PR #130, opened 2017-07-13
Fix some compiler warnings

TimMoore:
These mostly relate to Java generics and some edge cases we have encountered. This doesn't fix all of the warnings, however, so some follow up work is still required.

### DIFF
--- a/item-impl/src/test/java/com/example/auction/item/impl/ItemEntityTest.java
+++ b/item-impl/src/test/java/com/example/auction/item/impl/ItemEntityTest.java
@@ -274,7 +274,7 @@ public class ItemEntityTest {
 
     //  ---------------------------------------------------------------------------------------------------------
 
-    private <T> void expectEvents(Outcome<T, ?> outcome, T... expected) {
+    private void expectEvents(Outcome<?, ?> outcome, Object... expected) {
         if (!outcome.events().equals(Arrays.asList(expected))) {
             throw new AssertionError("Failed expectation. Expected [" + Arrays.asList(expected) + "] was not equal to [" + outcome.events() + "].");
         }
--- a/testkit/src/main/java/com/example/testkit/ReadSideTestDriver.java
+++ b/testkit/src/main/java/com/example/testkit/ReadSideTestDriver.java
@@ -10,6 +10,7 @@ import com.example.core.CompletionStageUtils;
 import com.google.inject.Injector;
 import com.google.inject.Singleton;
 import com.lightbend.lagom.javadsl.persistence.*;
+import com.lightbend.lagom.javadsl.persistence.ReadSideProcessor.ReadSideHandler;
 import org.pcollections.PSequence;
 
 import javax.inject.Inject;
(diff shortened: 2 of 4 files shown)


item-impl/src/test/java/com/example/auction/item/impl/ItemEntityTest.java
```diff
@@ -274,7 +274,7 @@ public void shouldReturnItemWithWinningBidAfterAuction() {
 
     //  ---------------------------------------------------------------------------------------------------------
 
-    private <T> void expectEvents(Outcome<T, ?> outcome, T... expected) {
+    private void expectEvents(Outcome<?, ?> outcome, Object... expected) {
```
TimMoore: Type parameters and varargs do not combine well in Java, and the original code here produces a warning. See http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ300

In this case, we don't actually care about the type of the expected events, since we are using them in `equals`.

testkit/src/main/java/com/example/testkit/ReadSideTestDriver.java
```diff
@@ -67,14 +68,15 @@ public ReadSideTestDriver(Injector injector, Materializer materializer) {
...
+    List<Pair<ReadSideHandler<?>, Offset>> list = processors.get(tag.eventType());
 
     if (list == null) {
       throw new RuntimeException("No processor registered for Event " + tag.eventType().getCanonicalName());
     }
 
-    List<CompletionStage<Done>> stages = list.stream().map(pHandlerOffset -> {
-          Flow<Pair<Event, Offset>, Done, ?> flow = ((ReadSideProcessor.ReadSideHandler<Event>) pHandlerOffset.first()).handle();
+    List<CompletionStage<?>> stages = list.stream().map(pHandlerOffset -> {
+      @SuppressWarnings("unchecked") ReadSideHandler<Event> handler = (ReadSideHandler<Event>) pHandlerOffset.first();
```
TimMoore: The only change here was to add an explicit `@SuppressWarnings("unchecked")` annotation, so we don't get a warning here. Unfortunately, there is no way to write this code in Java that does not produce a warning, but the class itself ensures the safety of this cast. Using the annotation here means we don't see the warning, but only for this specific variable.

All of the other changes in this file are just refactoring.

tools/src/main/java/com/example/core/CompletionStageUtils.java
```diff
@@ -28,11 +28,11 @@ private CompletionStageUtils() {
         };
     }
 
-    public static <T> CompletionStage<Done> doAll(CompletionStage<T>... stages) {
+    public static CompletionStage<Done> doAll(CompletionStage<?>... stages) {
         return doAll(Arrays.asList(stages));
     }
 
-    public static <T> CompletionStage<Done> doAll(List<CompletionStage<T>> stages) {
+    public static CompletionStage<Done> doAll(List<CompletionStage<?>> stages) {
```
TimMoore: Again, in this file we were getting warnings because of the combination of varargs and generics. In fact, we don't care about the result types of the completion stages, because these methods throw out the results and replace them with `Done` anyway.

TimMoore:
@yg-apaza & @lakhina have you had a chance to review this one? Thanks!

TimMoore:
Thanks for the reviews!

Issue #120, opened 2017-06-30
Replace ItemService.getUsers


TimMoore:
Right now, the `getUsers` service implementation is not ideal:

```java
        // Note this should never make production....
        return req -> currentIdsQuery.currentPersistenceIds()
                .filter(id -> id.startsWith("PUserEntity"))
                .mapAsync(4, id -> entityRef(id.substring(11)).ask(PUserCommand.GetPUser.INSTANCE))
                .filter(Optional::isPresent)
                .map(user -> Mappers.toApi(user.get()))
                .runWith(Sink.seq(), mat)
                .thenApply(TreePVector::from);
```

As you can tell from the comment, this code is a quick hack... not something that should be considered a good example of writing a Lagom service.

It has a few problems:

- It is slow. This requires loading each user entity into memory to query it for its details. In a multi-node cluster, this will require multiple calls across the network.
- It is fragile. It assumes too much about the format of the persistence ID string. See #115 for an example of the problems this causes
- It does not scale well. In a realistic system with a large number of users, this will return a very long list with no pagination.
- It is written in a very functional style that may be unfamiliar to many Java developers.

This service call is used by the web-gateway application to load a list of all users in the `Nav`. This is then used in turn for different purposes:

1. To create a list of users in the navigation bar, which allows you to log in as different users
2. To look up user details given a user ID, as an optimization of calling `userService.getUser` for each ID, since all of the user information was already loaded and cached locally for the navigation bar

These can be replaced by different means.

## Navigation bar

In the navigation bar, it might be better to show a bounded list of recently-created users. For example, we could put the ten most recently-created users in the list, rather than all of the users in the system. For most demo purposes, it will be exactly the same, since most people won't create that many users. However, if we decide to extend the demo to create users automatically (see #110 for example) then we might want to include a script that creates many more than ten users.

This is a perfect case for a read-side table in Lagom. This would look similar to the existing read-side processor in the item service and the one currently under review in the transaction service (#106). It would maintain a table of users and their creation date that could be queried by the service implementation in a new `getRecentUsers` service call. We would need to extend the `PUserCreated` event to carry a `createdAt` timestamp.

## User lookups

I think that using the navigation to look up user details, rather than making calls to `userService.getUser(id)`, could be considered a premature optimization. I suggest we undo that and stick with making the calls for now.

If we want to demonstrate optimizations in the future, we have a few options:

1. We could keep calling `getUser(id)` but maintain a local cache, so it doesn't have to be called as often
[...]

PR #131, opened 2017-07-13
Add readside processor to user

lakhina:
Phase one of #120 is complete. Now getUsers() provide you 10 users in nav list in with alphabatically least uuid


user-impl/src/main/java/com/example/auction/user/impl/PUser.java
```diff
@@ -2,6 +2,7 @@
 
 import com.lightbend.lagom.serialization.Jsonable;
 import lombok.Value;
+import lombok.experimental.Wither;
```
TimMoore: Is it used?

user-api/src/main/java/com/example/auction/user/api/UserService.java
```diff
@@ -1,31 +1,37 @@
...
         return named("user").withCalls(
                 pathCall("/api/user", this::createUser),
                 pathCall("/api/user/:id", this::getUser),
-                pathCall("/api/user", this::getUsers)
-        ).withPathParamSerializer(UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString));
+                pathCall("/api/user?pageNo&pageSize", this::getUsers)
+        ).withPathParamSerializer(
+                UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString)
+        ).withHeaderFilter(SecurityHeaderFilter.INSTANCE);
+      //  ).withPathParamSerializer(UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString));
```
TimMoore: The commented line can be deleted, since it is duplicated above.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,180 @@
...
+                return completedStatements(
+                        (List<BoundStatement>) insertUserCreator(user)
+                );
+            }
+            private BoundStatement insertUserCreator(PUser user) {
+                return insertUserStatement.bind(
+
+                        user.getId(),
+                        user.getEmail(),
+                        user.getName()
```
TimMoore: This doesn't look quite right. The statement you are binding to is this:

```
"INSERT INTO User(UserId, Name, email, PasswordHash) VALUES (?, ?, ?, ?)"
```

You need to bind the same number of parameters, in the same order they appear in the statement.

TimMoore: However, I would also say that you do not need to include the `PasswordHash` column in this table at all, since it is not used on the read side.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,180 @@
...
+                        )
+                        .thenApply(accept(s -> insertUserStatement = s ));
+            }
+
+
+
+
+            private CompletionStage<List<BoundStatement>> insertUser(PUser user) {
+                return completedStatements(
+                        (List<BoundStatement>) insertUserCreator(user)
```
TimMoore: This cast will fail: a `BoundStatement` cannot be cast to a `List<BoundStatement>`. You will need to wrap it in a list instead, such as `Arrays.asList(insertUserCreator(user))`.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            private CompletionStage<Done> createTables() {
+                return doAll(
+                        session.executeCreateTable(
+                                "CREATE TABLE IF NOT EXISTS UserInfo (" +
+                                        "UserId UUID PRIMARY KEY, " +
+                                        "Name text, "+
+                                        "email text, " +
+                                        "PasswordHash text "+
```
TimMoore: As mentioned in another comment, there's no need for the `PasswordHash` column, since it is never read.

user-api/src/main/java/com/example/auction/user/api/UserService.java
```diff
@@ -1,31 +1,37 @@
...
     default Descriptor descriptor() {
         return named("user").withCalls(
                 pathCall("/api/user", this::createUser),
                 pathCall("/api/user/:id", this::getUser),
-                pathCall("/api/user", this::getUsers)
-        ).withPathParamSerializer(UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString));
+                pathCall("/api/user?pageNo&pageSize", this::getUsers)
+        ).withPathParamSerializer(
+                UUID.class, PathParamSerializers.required("UUID", UUID::fromString, UUID::toString)
+        ).withHeaderFilter(SecurityHeaderFilter.INSTANCE);
```
TimMoore: What's the reason for adding the `SecurityHeaderFilter`? I don't see it used anywhere.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,180 @@
...
+
+
+            }
+
+
+
+            private CompletionStage<Done> prepareInsertUserStatement() {
+
+                return session.
+                        prepare("INSERT INTO User(" +
```
TimMoore: The name of the table here ("User") does not match the table created above ("UserInfo").

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,180 @@
...
+            private BoundStatement insertUserCreator(PUser user) {
+                return insertUserStatement.bind(
+
+                        user.getId(),
+                        user.getEmail(),
+                        user.getName()
+                );
+            }
+            private CompletionStage<Optional<Row>> selectUser(UUID UserId) {
+                return session.selectOne("SELECT * FROM UserCreator WHERE UserId = ?", UserId);
```
TimMoore: This table name also does not match either of the other two named used.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,181 @@
...
+
+
+        private CompletionStage<Done> prepareInsertUserStatement() {
+
+            return session.
+                    prepare("INSERT INTO UserInfo(" +
+
+                            "UserId, " +
+                            "Name, " +
+                            "email, " +
```
yg-apaza: The [actual error](https://travis-ci.org/lagom/online-auction-java/builds/253984001#L3087): `no viable alternative at input ')' (...(UserId, Name, email, [)]...)`, it's because there is an extra comma

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,181 @@
...
+
+                            ") VALUES (" +
+                            "?, " + // UserId
+                            "?, " + // Name
+                            "?, " + // email
```
yg-apaza: Here is the same extra comma, I think it's better to remove in order to avoid confusion

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,181 @@
...
+            return PUserEvent.TAG.allTags();
+        }
+
+        private CompletionStage<Done> createTables() {
+            return doAll(
+                    session.executeCreateTable(
+                            "CREATE TABLE IF NOT EXISTS UserInfo (" +
+                                    "UserId UUID PRIMARY KEY, " +
+                                    "Name text, " +
+                                    "email text, " +
```
TimMoore: I'm not sure what happened to the comment from @yg-apaza pointing out the same, but there is a syntax error here: trailing comma on the last column.

lakhina: @TimMoore  @yg-apaza  Thanks. I will fix it.

user-api/src/main/java/com/example/auction/user/api/User.java
```diff
@@ -11,12 +11,14 @@
 public final class User {
 
     private final UUID id;
+    private final UUID creatorId;
     private final String name;
     private final String email;
 
     @JsonCreator
-    public User(@JsonProperty("id") UUID id, @JsonProperty("name") String name, @JsonProperty("email") String email) {
+    public User(@JsonProperty("id") UUID id, @JsonProperty("creatorId") UUID creatorId , @JsonProperty("name") String name, @JsonProperty("email") String email) {
```
TimMoore: I think we want it to be a timestamp (maybe called `createdAt`) rather than a UUID.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,192 @@
...
+                    Arrays.asList(insertUserCreator(user))
+
+            );
+        }
+
+        private BoundStatement insertUserCreator(PUser user) {
+            return insertUserStatement.bind(
+
+                    user.getId(),
+                    user.getCreatedAt(),
```
TimMoore: In order to bind an `Instant` to a Cassandra query, you can use the optional [`InstantCodec`](http://docs.datastax.com/en/developer/java-driver/3.1/manual/custom_codecs/extras/#temporal-types) provided by the Java Cassandra driver.

You can see an example of how to register a codec in [`ItemRepository`](https://github.com/lagom/online-auction-java/blob/5db2bda15166ce9a632905ecf75aeed7f3cd5b07/item-impl/src/main/java/com/example/auction/item/impl/ItemRepository.java#L174) (and there is a similar example in `TransactionRepository`).

web-gateway/src/main/resources/application.conf
```diff
@@ -1,5 +1,5 @@
 play.http.secret.key = "somesecret"
 online-auction.instruction.show = true
-
+lagom.persistence.read-side.global-prepare-timeout = 60s
```
TimMoore: If this doesn't fix the problem for you, then it should probably be removed. We'll need to find some other solution.

TimMoore:
@lakhina the latest build has a new error:

```
 Caused by: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: com.datastax.driver.core.exceptions.SyntaxError: line 1:107 mismatched input '(' expecting ')' (..., email text PRIMARY KEY [(]createdAt...)
```

This indicates a syntax error in your CQL statement. Have a look at the stack trace to find where.

The error "Couldn't register the processor on the testkit" is a generic wrapper exception that could have a variety of different causes. It's important to always read the nested exceptions to find the root cause of the error.

rcavalcanti:
@lakhina I have been trying to trace why the build is failing. 

Most of the times it fails with a timeout on the actor system when it tries to create the tables. I could find the cause of that yet. I will keep searching. 

However, from time to time it succeeds and when it happens we get other errors on the `UserRepositoryTest` (see my comments there)

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -0,0 +1,199 @@
...
+
+                            ")"
+                    )
+                    .thenApply(accept(s -> insertUserStatement = s));
+        }
+
+
+        private CompletionStage<List<BoundStatement>> insertUser(PUser user) {
+            return completedStatements(
+                    Arrays.asList(insertUserCreator(user))
```
rcavalcanti: Here we can remove the `Arrays.asList` because there is an overloaded `completedStatements` that gets a varargs.

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test
+    public void shouldPaginateUserRetrieval() throws InterruptedException, ExecutionException, TimeoutException {
+
+        for (int i = 0; i < 25; i++) {
+
+            feed(new PUserEvent.PUserCreated(buildFixture(createdAt, i)));
+        }
+
+        PaginatedSequence<User> createdUsers = Await.result(userRepository.getUsers(1, 10));
+        assertEquals(25, createdUsers.getCount());
```
rcavalcanti: Because we insert a `User` on previous test, the count here will be 26. 

We have two options: 
* make sure table is cleared before each test (costly)
* cache the current count before insert the new `Users` and compare with it.

```
int size = 25
int initialCount = createdUsers.getCount();

// create users here

assertEquals(initialCount + size, createdUsers.getCount());
```

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals(10, createdUsers.getItems().size());
+        // default ordering is time DESC so page 2 of size 10 over a set of 25 returns item ids 5-14. On that seq, the fifth item is id=10
+        assertEquals("user10", createdUsers.getItems().get(4).getName());
```
rcavalcanti: I'm getting `admin4` instead. Not "user10"

lakhina:
@rcavalcanti  Thanks for the help. I have made the suggested changes but still the same error is persisting.

rcavalcanti:
@lakhina the fix I suggested won't fix the build problem (see [previous comment](https://github.com/lagom/online-auction-java/pull/131#issuecomment-318509437)).

What I observed is that when it happens to pass the stage of table creation, it does fail because of errors in the test itself. 

Can you push your last changes? Thanks

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -56,8 +56,12 @@ public static void afterAll() {
     private final String email = "admin@gmail.com";
     private final String password = PUserCommand.hashPassword("admin");
     private final Instant createdAt = Instant.now();
-
+    int size = 25;
+    PaginatedSequence<User> createdUsers = Await.result(userRepository.getUsers(1, 10));
     private final PUser userCreated = new PUser(userId, createdAt, name, email, password);
+    int initialCount = createdUsers.getCount();
```
rcavalcanti: I think it's better to have it inside the test method instead. 

`size` is the number of users you plan to create inside `shouldPaginateUserRetrieval` method. 
Preferably, we should use this value in the for loop as well.

```
for (int i = 0; i < size; i++) { ... }
```

The `initialCount` as well. This call will happen on class initialisation. At that point, the initial count is 0. 

My suggestion was more something like the following:

```
 @Test
public void shouldPaginateUserRetrieval() throws InterruptedException, ExecutionException, TimeoutException {

    PaginatedSequence<User> usersCreatedBefore = Await.result(userRepository.getUsers(1, 10));
    int initialCount = usersCreatedBefore.getCount();
    int size = 25;

    for (int i = 0; i < size; i++) {
        feed(new PUserEvent.PUserCreated(buildFixture(createdAt, i)));
    }

    PaginatedSequence<User> createdUsers = Await.result(userRepository.getUsers(1, 10));
    assertEquals(size + initialCount, createdUsers.getCount());
}
```

rcavalcanti: Looking back to the code I just pasted here, it is even better to call `userRepository.countUsers()` instead of `userRepository.getUsers(1, 10)`.  Certainly for the first call where we are only interested in the total current count.

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -87,11 +91,13 @@ public void shouldPaginateUserRetrieval() throws InterruptedException, Execution
...
 
+
         PaginatedSequence<User> createdUsers = Await.result(userRepository.getUsers(1, 10));
-        assertEquals(25, createdUsers.getCount());
+        assertEquals(initialCount + size, createdUsers.getCount());
+
         assertEquals(10, createdUsers.getItems().size());
         // default ordering is time DESC so page 2 of size 10 over a set of 25 returns item ids 5-14. On that seq, the fifth item is id=10
-        assertEquals("user10", createdUsers.getItems().get(4).getName());
+        assertEquals("admin4", createdUsers.getItems().get(4).getName());
```
rcavalcanti: It seems that I was wrong about this one. I'm getting random values here. 
I've seen the following values so far:
`admin4`
`admin10`
`admin22`
`admin18`

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
rcavalcanti: This is not a syntax error, on the contrary.

If I remove the trailing comma I get a syntax error from the cassandra driver.
`ItemRepository` also have a comma separating a the last column and the definition of a compound primary key.

There is two slightly different syntax.
```
CREATE TABLE demo (
 foo text PRIMARY KEY,
 bar text
)
```
No comma between `foo` column definition and `PRIMARY KEY` as we are marking `foo` as the primary key.

```
CREATE TABLE demo (
 foo text,
 bar text,
 PRIMARY KEY(foo, bar)
)
```
In the above example we do need a comma because we are defining a compound PK. In that case, "PRIMARY KEY" is not part of the column definition and therefore a comma is needed

rcavalcanti: @TimMoore, I think that @yg-apaza's disappeared because she may have realised it and removed it.

lakhina: @rcavalcanti  I have changed this query now.

rcavalcanti: hmm, seems that I had already checked out the one with the compound key that made the comments here obsolete. Sorry for the noisy on the line.

lakhina:
@rcavalcanti  I have committed the changes. Please have a look.

user-impl/src/main/java/Module.java
```diff
@@ -1,12 +1,13 @@
-package com.example.auction.user.impl;
-
 import com.example.auction.user.api.UserService;
+import com.example.auction.user.impl.UserRepository;
+import com.example.auction.user.impl.UserServiceImpl;
 import com.google.inject.AbstractModule;
 import com.lightbend.lagom.javadsl.server.ServiceGuiceSupport;
 
-public class UserModule extends AbstractModule implements ServiceGuiceSupport {
+public class Module extends AbstractModule implements ServiceGuiceSupport {
```
rcavalcanti: I could finally detect the origin of the broken build. I don't have yet full understand how does it happens, but it's related with the order of wiring of Guice modules. 

It seems that when using a non-standard location for the Guice module, the bit that will make the akka cluster to join itself is affected by the bootstrap of cassandra. Probably by the loading of the Cassandra Module.

What I have observed is that sometimes (1 over 3 times), the cluster joining (`JoinClusterImpl.join`) is only called after some Cassandra setup. 

```
2017-07-31 11:17:32,128 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - Starting up...
2017-07-31 11:17:32,214 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - Registered cluster JMX MBean [akka:type=Cluster]
2017-07-31 11:17:32,215 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - Started up successfully
2017-07-31 11:17:32,236 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - No seed-nodes configured, manual cluster join required
...
2017-07-31 11:17:35,577 [DEBUG] from com.datastax.driver.core.ControlConnection - [Control connection] Refreshing schema for servicetest_170731111724385
2017-07-31 11:17:39,781 [INFO] from org.apache.cassandra.auth.CassandraRoleManager - Created default superuser role 'cassandra'

// JoinClusterImpl.join is called here
2017-07-31 11:17:54,137 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] is JOINING, roles []
2017-07-31 11:17:54,150 [INFO] from akka.cluster.Cluster(akka://ServiceTest_170731111724385) - Cluster Node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] - Leader is moving node [akka.tcp://ServiceTest_170731111724385@127.0.0.1:53609] to [Up]
```

When this happens, the `UserRepositoryTest` fails because it depends on async call to `ClusterStartupTaskActor` in order to get an instance of `CassandraOffsetDao` which is required by the cassandra read side.

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+        int size = 25;
+
+        for (int i = 0; i < size; i++) {
+            feed(new PUserEvent.PUserCreated(buildFixture(createdAt, i)));
+        }
+
+        PaginatedSequence<User> createdUsers = Await.result(userRepository.getUsers(1, 10));
+        assertEquals(size + initialCount, createdUsers.getCount());
+        assertEquals(10, createdUsers.getItems().size());
+        assertEquals("admin11", createdUsers.getItems().get(4).getName());
```
rcavalcanti: This one will randomly fail because the ordering is not guaranteed. 

`UserRepository.selectUsers` must include a `order by createdAt` to make this call deterministic.

user-impl/src/test/java/com/example/auction/user/impl/UserRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals(1, users.getCount());
+        User expected = new User(userId, createdAt, name, email);
+        assertEquals(expected, users.getItems().get(0));
+    }
+
+    @Test
+    public void shouldPaginateUserRetrieval() throws InterruptedException, ExecutionException, TimeoutException {
+
+        PaginatedSequence<User> usersCreatedBefore = Await.result(userRepository.getUsers(1, 10));
+        int initialCount = usersCreatedBefore.getCount();
```
rcavalcanti: If we make `UserRepository.countUsers()` public, we can calculate the `initialCount` in one shot without the end fetch results in-memory.

```int initialCount = Await.result(userRepository.countUsers());```

user-impl/src/main/java/Module.java
TimMoore: Wow, that is some very good detective work! How strange.

We have noticed some other issues related to this in the past, for example, https://github.com/lagom/lagom/issues/568. That issue pertains to JDBC rather than Cassandra, but I think it's similar in that the underlying issue is a lack of coordination in the initialization order. All of our `ClusterStartupTask` implementations kick off immediately as soon as they are defined.

@rcavalcanti: could you record your findings in a lagom/lagom issue? We should really fix this upstream.

TimMoore:
With the latest changes from @rcavalcanti, it looks like the build failure is now a legitimate test failure (should be reproducible locally) rather than a problem with the framework.

Issue #146, opened 2017-08-11
Use gracefulShutdown as described in Lagom docs.


ignasi35:
From lagom [docs](https://www.lagomframework.com/documentation/1.3.x/java/Cluster.html#Leaving):

> When using Persistent Entities you can use PersistentEntityRegistry.gracefulShutdown to stop the persistent entities and leave the cluster in a graceful way. This is not mandatory but it can be good when you are doing a controlled shutdown of a service node. It will reduce the number of lost in-flight messages during the failover to another node.

Some services (probably bid, item and transaction) should use gracefulShutdown.

PR #150, opened 2017-08-16
Use gracefulShutdown in transaction service

yg-apaza:
Addresses part of #146

### DIFF
--- a/transaction-impl/src/main/java/Module.java
+++ b/transaction-impl/src/main/java/Module.java
@@ -1,8 +1,14 @@
-import com.example.auction.transaction.impl.TransactionServiceImpl;
 import com.example.auction.item.api.ItemService;
 import com.example.auction.transaction.api.TransactionService;
+import com.example.auction.transaction.impl.TransactionServiceImpl;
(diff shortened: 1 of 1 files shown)


transaction-impl/src/main/java/Module.java
```diff
@@ -12,7 +18,15 @@
...
         bindClient(ItemService.class);
+
+        bind(TransactionServiceLifecycle.class).asEagerSingleton();
+    }
+
+    private static class TransactionServiceLifecycle {
+        @Inject
+        public TransactionServiceLifecycle(ApplicationLifecycle applicationLifecycle, PersistentEntityRegistry persistentEntityRegistry) {
+            applicationLifecycle.addStopHook(() -> persistentEntityRegistry.gracefulShutdown(FiniteDuration.create(5, TimeUnit.SECONDS)));
+        }
```
ignasi35: That's a great approach. We usually add the `addStopHook` in the constructor of the `XxxServiceImpl` but this pattern makes things cleaner IMHO (keeps all wiring and shutdown in a single place).

PR #151, opened 2017-08-17
Add feature on Transaction Service: Submit payment status (approved/rejected)

yg-apaza:
- [x] Add entity tests, service tests, repository tests
- [x] Update `TransactionEntity`, `TransactionServiceImpl` and `TransactionRepository`

### DIFF
--- a/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
+++ b/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
@@ -61,6 +61,7 @@ public interface TransactionService extends Service {
                 pathCall("/api/transaction/:id/deliveryprice", this::setDeliveryPrice),
                 pathCall("/api/transaction/:id/approvedelivery", this::approveDeliveryDetails),
                 pathCall("/api/transaction/:id/paymentdetails", this::submitPaymentDetails),
+                pathCall("/api/transaction/:id/paymentstatus", this::submitPaymentStatus),
                 pathCall("/api/transaction/:id", this::getTransaction),
                 pathCall("/api/transaction?status&pageNo&pageSize", this::getTransactionsForUser)
         ).withPathParamSerializer(
--- a/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
+++ b/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
@@ -36,7 +36,7 @@ public interface TransactionService extends Service {
 
     ServiceCall<PaymentInfo, Done> submitPaymentDetails(UUID itemId);
 
-    ServiceCall<Boolean, Done> submitPaymentStatus(UUID itemId);
+    ServiceCall<PaymentInfoStatus, Done> submitPaymentStatus(UUID itemId);
 
     //ServiceCall<NotUsed, Done> dispatchItem(UUID itemId);
 
@@ -70,5 +70,4 @@ public interface TransactionService extends Service {
                 TransactionInfoStatus.class, PathParamSerializers.required("TransactionInfoStatus", TransactionInfoStatus::valueOf, TransactionInfoStatus::toString)
         ).withHeaderFilter(SecurityHeaderFilter.INSTANCE);
     }
-
 }
--- a/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
+++ b/transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
@@ -36,6 +36,8 @@ public interface TransactionService extends Service {
 
     ServiceCall<PaymentInfo, Done> submitPaymentDetails(UUID itemId);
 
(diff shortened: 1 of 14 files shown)


transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
```diff
@@ -133,6 +135,32 @@ private Behavior paymentPending(TransactionState state) {
...
+                throw new Forbidden("Only the item creator can approve or reject payment");
+        });
+
+        builder.setEventHandlerChangingBehavior(PaymentApproved.class, evt ->
+                paymentConfirmed(state().withStatus(TransactionStatus.PAYMENT_CONFIRMED))
+        );
+
+        builder.setEventHandlerChangingBehavior(PaymentRejected.class, evt ->
+                paymentConfirmed(state().withStatus(TransactionStatus.PAYMENT_PENDING))
+        );
```
ignasi35: Is this behavioural change correct? If the payment was rejected I wouldn't expect the behaviour to be `paymentConfirmed`

transaction-api/src/main/java/com/example/auction/transaction/api/TransactionService.java
```diff
@@ -36,6 +36,8 @@
 
     ServiceCall<PaymentInfo, Done> submitPaymentDetails(UUID itemId);
 
+    ServiceCall<Boolean, Done> submitPaymentStatus(UUID itemId);
```
ignasi35: How about using an enum as a pyaload? It would allow extending in the future while keeping backwards compatibility. Boolean is restricting us to `yes/no` while an enum can be extended.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
ignasi35: I reviewed the class as a whole and I think the method should be `transactionFinished` and the status `PAYMENT_CONFIRMED` and `PAYMENT_REJECTED`. I'm avoiding words like `confirmed` or `completed` that give a sense of success. IIUC, this is a transition into a final state that users can't get out of and whic contains a status of success or failure.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionServiceImpl.java
```diff
@@ -86,6 +88,14 @@ public TransactionServiceImpl(PersistentEntityRegistry registry, ItemService ite
     }
 
     @Override
+    public ServiceCall<Boolean, Done> submitPaymentStatus(UUID itemId) {
+        return authenticated(userId -> request -> {
+            SubmitPaymentStatus submitPaymentStatus = new SubmitPaymentStatus(userId, request);
+            return entityRef(itemId).ask(submitPaymentStatus);
```
ignasi35: If request wasn't a `boolean` we would be able to map the values from the request into internal `TransactionStatus` here.

transaction-impl/src/main/java/com/example/auction/transaction/impl/TransactionEntity.java
TimMoore: @ignasi35, @yg-apaza and I discussed this on our call... here's what I was thinking: if the payment is rejected, that doesn't necessary end the whole thing. The seller and buyer can try again to negotiate payment, so it should return to the `paymentPending` behavior.

Eventually, we should add the option of the seller cancelling the transaction as a manual step.

TimMoore: So, to summarize, this code does still need to change, to call `paymentPending` instead of `paymentConfirmed`.

ignasi35: That makes sense. Thanks ;-)

yg-apaza:
@TimMoore @ignasi35 I made the suggested changes, please review

ignasi35:
LGTM, @TimMoore do you have any further comment?

transaction-impl/src/main/java/com/example/auction/transaction/impl/PaymentStatus.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.example.auction.transaction.api.PaymentInfoStatus;
+
+public enum PaymentStatus {
+    APPROVED(PaymentInfoStatus.APPROVED), //{
+//        @Override
+        //PaymentInfoStatus toPaymentInfoStatus() {
+  //          return PaymentInfoStatus.APPROVED;
+//        }
+//    },
```
TimMoore: Can you delete the commented code?

yg-apaza:
Yes, I'll do the UI part in another pull request

yg-apaza:
I updated the title, how about now @TimMoore ?

TimMoore:
Thanks!

PR #152, opened 2017-08-23
Add login form in User

lakhina:


user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -61,11 +62,34 @@ public UserServiceImpl(PersistentEntityRegistry registry, UserRepository userRep
...
         return entityRef(id.toString());
     }
 
     private PersistentEntityRef<PUserCommand> entityRef(String id) {
         return registry.refFor(PUserEntity.class, id);
     }
+
+    private PersistentEntityRef<PUserCommand> credRef(String email) {
+        return registry.refFor(PUserEntity.class, email);
+    }
```
ignasi35: I'm not sure what is the intent of this but you can't have two instances for the same User (a first instance referenced by the `userId` and a second instance referenced by the `email`).

```
     private PersistentEntityRef<PUserCommand> entityRef(String id) {
         return registry.refFor(PUserEntity.class, id);
     }
    private PersistentEntityRef<PUserCommand> credRef(String email) {
        return registry.refFor(PUserEntity.class, email);
    }
```

web-gateway/src/main/java/controllers/UserController.java
```diff
@@ -61,4 +61,29 @@ public Result currentUser(String userId) {
...
+    public CompletionStage<Result> loginUser() {
+        Http.Context ctx = ctx();
+        return withUser(ctx, userId ->
+                loadNav(userId).thenCompose(nav -> {
+                    Form<LoginForm> form = formFactory.form(LoginForm.class).bindFromRequest(ctx.request());
+                    if (form.hasErrors()) {
+                        return CompletableFuture.completedFuture(ok(views.html.login.render(showInlineInstruction, form, nav)));
+                    }
+
+                    return userService.login().invoke(new UserRegistration(nav.getUser().get().getName(), form.get().getEmail(), form.get().getPassword())).thenApply(id -> {
```
ignasi35: @lakhina I don't think the `login` operation should reuse the `UserRegistration`. It contains a `name` field which is not available: because `nav.getUser().get().getName()` can't be resolved since there's no logged in user.

lakhina:
@ignasi35  Please review

lakhina:
@TimMoore @ignasi35 I have made the suggested changes. Please review

user-impl/src/main/java/com/example/auction/user/impl/PUserEntity.java
```diff
@@ -53,4 +54,21 @@ private Behavior notCreated() {
...
+        return (hashed_password);
+    }
+
+    public static boolean checkPassword(String password_plaintext, String stored_hash) {
+        boolean password_verified = false;
+
+        if (null == stored_hash)
+            throw new java.lang.IllegalArgumentException("Invalid hash provided for comparison");
+
+        password_verified = BCrypt.checkpw(password_plaintext, stored_hash);
```
rcavalcanti: this could be replaced by 
```
return BCrypt.checkpw(password_plaintext, stored_hash);
```
and the `password_verified` var removed. 

(btw, the same in `hashPassword`)

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -79,6 +81,19 @@ public UserRepository(CassandraSession session, ReadSide readSide) {
...
+
+        return session
+            .selectOne(
+                "SELECT * FROM UserInfo  WHERE email = ? " ,
+                   email
+            )
+
+            .thenApply(rows -> rows.map(UserRepository::convertUserSummary).get())
+            .thenApply(UserSummary -> UserSummary.getId());
+
```
rcavalcanti: Good visual formatting. Each operation on a different line helps quick scanning of code and understanding what it's doing. 👍

PR #153, opened 2017-08-25
Add logout implementation

lakhina:


user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -42,6 +43,16 @@ public UserServiceImpl(PersistentEntityRegistry registry, UserRepository userRep
...
 
     @Override
+    public ServiceCall<NotUsed, Done> logout(UUID userId) {
+        return req -> {
+            return entityRef(userId).ask(PUserCommand.GetPUser.INSTANCE)
+                .thenApply(maybeCredential-> {
+                    return Done.getInstance();
+                });
+        };
+    }
```
TimMoore: I don't think it's necessary to have a `logout` call in the `UserService`... this method doesn't really do anything. It can be handled entirely in the web gateway.

web-gateway/src/main/java/controllers/UserController.java
```diff
@@ -57,6 +57,20 @@ public UserController(Configuration config, MessagesApi messagesApi, UserService
         );
     }
 
+    public CompletionStage<Result> logoutUser() {
+        Http.Context ctx = ctx();
```
TimMoore: I think this method can be reduced to just two lines and still do everything that this one does:

```java
ctx().session().clear();
return ok(views.html.logout.render(nav));
```

lakhina:
@TimMoore @ignasi35  I have made the suggested changes. Please review

web-gateway/src/main/java/controllers/UserController.java
ignasi35: @TimMoore I think `nav` would be unknown. There's an intermediate solution between your two options:

```java
    public CompletionStage<Result> logoutUser() {
        return loadNav(Optional.empty()).thenApply(nav -> {
            ctx().session().clear();
            return ok(views.html.logout.render(nav));
        });
    }
```

web-gateway/src/main/twirl/views/logout.scala.html
```diff
@@ -0,0 +1,5 @@
+@(implicit nav: Nav)
+
+@main("Logout") {
+  <h3>You have been successfully logged out.</h3>
+}
```
ignasi35: Do we need an extra page? Could we go redirect to the main screen with a cleared `ctx`?

lakhina: @ignasi35  I have now redirected to the index page

lakhina:
@ignasi35 @TimMoore Please review

ignasi35:
Looks great! Thanks @lakhina

PR #167, opened 2018-03-06
Bumps to Lagom 1.4.1. Fixes some ExCtx issues

ignasi35:
Bumps Lagom to 1.4.1.

While testing I noticed a couple of controllers failed because the CompletionStage composition required the explicit ExecutionContext.

### DIFF
--- a/web-gateway/src/main/java/controllers/ItemController.java
+++ b/web-gateway/src/main/java/controllers/ItemController.java
@@ -170,54 +170,55 @@ public class ItemController extends AbstractController {
     }
 
     private CompletionStage<Result> doGetItem(Http.Context ctx, String itemId, Form<BidForm> bidForm) {
-        return requireUser(ctx, user -> loadNav(user).thenComposeAsync(nav -> {
+        return requireUser(ctx, user -> loadNav(user)
+            .thenComposeAsync(nav -> {
 
-            UUID itemUuid = UUID.fromString(itemId);
-            CompletionStage<Item> itemFuture = itemService.getItem(itemUuid)
+                UUID itemUuid = UUID.fromString(itemId);
+                CompletionStage<Item> itemFuture = itemService.getItem(itemUuid)
                     .handleRequestHeader(authenticate(user)).invoke();
-            CompletionStage<PSequence<Bid>> bidHistoryFuture = bidService.getBids(itemUuid)
+                CompletionStage<PSequence<Bid>> bidHistoryFuture = bidService.getBids(itemUuid)
                     .handleRequestHeader(authenticate(user)).invoke();
-            return bidHistoryFuture.thenComposeAsync(bidHistory ->
+                return bidHistoryFuture.thenComposeAsync(bidHistory ->
 
-               itemFuture.thenComposeAsync(item1 ->
-                     {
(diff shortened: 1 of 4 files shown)


web-gateway/src/main/java/controllers/ItemController.java
```diff
@@ -170,54 +170,55 @@ private ItemData fromForm(ItemForm itemForm) {
...
+
+                                    Currency currency = Currency.valueOf(item.getItemData().getCurrencyId());
+
+                                    Optional<BidResult> bidResult = loadBidResult(ctx.flash());
+
+                                    return ok(views.html.item.render(showInlineInstruction, item, bidForm,
+                                        anonymizeBids(user, currency, bidHistory), user, currency, seller, winner, currentBidMaximum, bidResult, (Nav) nav));
+                                }, ec.current());
+                        }, ec.current())
+                    , ec.current());
```
ignasi35: This huge change is mostly formatting. The interesting bit are the last two `ec.current()` added.

web-gateway/src/main/java/controllers/UserController.java
```diff
@@ -43,30 +43,31 @@ public UserController(Configuration config, MessagesApi messagesApi, UserService
...
     public CompletionStage<Result> logoutUser() {
-        return loadNav(Optional.empty()).thenApply(nav -> {
-            ctx().session().clear();
-            return ok(views.html.index.render(nav));
-        });
+        return loadNav(Optional.empty())
+            .thenApplyAsync(nav -> {
+                ctx().session().clear();
+                return ok(views.html.index.render(nav));
+            }, httpExecutionContext.current());
```
ignasi35: All this change is formatting and a couple of `ec.current()` added in seldom methods.

ignasi35:
I've done a rather extensive manual test of the (including search features). Keeping an eye on terminal logs too.

With the changes suggested in this PR everything seems to work.

Hmm, not sure I tested the login operation (I did test signup and logout, though). 🤔

web-gateway/src/main/java/controllers/ItemController.java
TimMoore: I'm a little surprised that it needs to be added in so many places. I thought it would only be the `render` call that needs to run in that EC.

This whole thing could stand some refactoring I think... later

TimMoore:
Easier to review with the secret "ignore whitespace" URL: https://github.com/lagom/online-auction-java/pull/167/files?w=1

TimMoore:
Login page is in fact broken

TimMoore:
It looks like the entire login function is broken. I suggest that we leave that out of this pull request.

TimMoore:
I pushed my WIP changes to the login controller, but the back-end service also needs to be fixed.

TimMoore:
Tests repeatedly timed out, so I increased the timeout duration.

ignasi35:
Thanks @TimMoore !

web-gateway/src/main/java/controllers/ItemController.java
ignasi35: > This whole thing could stand some refactoring I think... later

Oh, it does!

PR #169, opened 2018-03-07
Fixes login for happy path

ignasi35:
This PR replaces the way the userId is computed by a derived alg that canonicalizes the e-mail and uses it's `byte[]` to build the UUID. It's not great but it works for the time being.

The canonicalization uses a `toLowerCase` so that username is not case sensitive on login. Password still is case sensitive.

This PR removes the offending Cassandra query that serached by `email = ?`.

This PR doesn't fix the cosmetic issues (the login page should be displayed instead of a default play screen)  or data leaks in case of failed login attempts (`wrong e-mail` uses a different error message than `valid email but wrong password`).

Lots of code formatting.

### DIFF
--- a/item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
+++ b/item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
@@ -41,7 +41,7 @@ import static org.junit.Assert.assertEquals;
  */
 public class ItemServiceImplIntegrationTest {
 
-    private final static Setup setup = defaultSetup().withCassandra(true)
+    private final static Setup setup = defaultSetup().withCassandra()
             .configureBuilder(b ->
                     // by default, cassandra-query-journal delays propagation of events by 10sec. In test we're using
                     // a 1 node cluster so this delay is not necessary.
--- a/item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
+++ b/item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
@@ -41,7 +41,7 @@ import static org.junit.Assert.assertEquals;
  */
 public class ItemServiceImplIntegrationTest {
 
-    private final static Setup setup = defaultSetup().withCassandra()
+    private final static Setup setup = defaultSetup().withCassandra(true)
             .configureBuilder(b ->
                     // by default, cassandra-query-journal delays propagation of events by 10sec. In test we're using
                     // a 1 node cluster so this delay is not necessary.
--- a/user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
+++ b/user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
@@ -44,63 +44,45 @@ public class UserRepository {
 
     CompletionStage<PaginatedSequence<User>> getUsers(int page, int pageSize) {
         return countUsers()
-                .thenCompose(
-                        count -> {
-                            int offset = page * pageSize;
-                            int limit = (page + 1) * pageSize;
-                            CompletionStage<PSequence<User>> Users = offset > count ?
-                                    CompletableFuture.completedFuture(TreePVector.empty()) :
-                                    selectUsers(offset, limit);
-                            return Users.thenApply(seq -> new PaginatedSequence<>(seq, page, pageSize, count));
-                        }
-                );
+            .thenCompose(
+                count -> {
+                    int offset = page * pageSize;
+                    int limit = (page + 1) * pageSize;
+                    CompletionStage<PSequence<User>> Users = offset > count ?
+                        CompletableFuture.completedFuture(TreePVector.empty()) :
+                        selectUsers(offset, limit);
+                    return Users.thenApply(seq -> new PaginatedSequence<>(seq, page, pageSize, count));
+                }
+            );
     }
 
     public CompletionStage<Integer> countUsers() {
         return session
-                .selectOne(
-                        "SELECT COUNT(*) FROM UserInfo  "
-
-                )
-                .thenApply(row -> (int) row.get().getLong("count"));
+            .selectOne(
+                "SELECT COUNT(*) FROM UserInfo  "
+            )
+            .thenApply(row -> (int) row.get().getLong("count"));
     }
 
     private CompletionStage<PSequence<User>> selectUsers(long offset, int limit) {
-
         return session
-                .selectAll(
-                        "SELECT * FROM UserInfo  " +
-                                "LIMIT ?",
-
-                        limit
-                )
-                .thenApply(List::stream)
-                .thenApply(rows -> rows.skip(offset))
-                .thenApply(rows -> rows.map(UserRepository::convertUserSummary))
-                .thenApply(UserSummaries -> UserSummaries.collect(Collectors.toList()))
-                .thenApply(TreePVector::from);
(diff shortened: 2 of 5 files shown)


user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
```diff
@@ -44,63 +44,45 @@ public UserRepository(CassandraSession session, ReadSide readSide) {
...
-
-            .thenApply(rows -> rows.map(UserRepository::convertUserSummary).get())
-            .thenApply(UserSummary -> UserSummary.getId());
-
-
+            .thenApply(List::stream)
+            .thenApply(rows -> rows.skip(offset))
+            .thenApply(rows -> rows.map(UserRepository::convertUserSummary))
+            .thenApply(UserSummaries -> UserSummaries.collect(Collectors.toList()))
+            .thenApply(TreePVector::from);
```
ignasi35: This whole block above is a bit messy on the GH UI. There are two changes: `selectUsers` was formatted and `getUserIdByEmail` was removed.

(using `?w=1` in the URL won't let me add review comments :-( )

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -26,37 +26,35 @@
...
+            PUser createdUser = new PUser(uuid, user.getName(), user.getEmail(), password);
             return entityRef(uuid)
                 .ask(new PUserCommand.CreatePUser(user.getName(), user.getEmail(), password))
                 .thenApply(done -> Mappers.toApi(Optional.ofNullable(createdUser)));
         };
     }
 
+    private UUID getUUIDFromEmail(String email) {
+        return UUID.nameUUIDFromBytes(email.toLowerCase().getBytes());
+    }
```
ignasi35: This is the major change on this PR: `userId` is derived from the `email`.  This is a first step into making the `userId` the actual user `email` but that would have taken me a bit longer.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -1,40 +1,91 @@
...
-        });
+        String name = "admin";
+        String email = "admin@gmail.com";
+        String password = "admin";
+        userService = server.client(UserService.class);
+        UserRegistration userRegistration = new UserRegistration(name, email, password);
+        User createdUser = createNewUser(userRegistration);
+        assertEquals(name, createdUser.getName());
+        assertEquals(email, createdUser.getEmail());
+    }
```
ignasi35: Same test as before but some formatting

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -1,40 +1,91 @@
...
+    // TODO: replace with Forbidden.class
+    @Test(expected = NotFound.class)
+    public void shouldDenyLoggingInWithInalidCredentials() throws Throwable {
+        try {
+            Await.result(userService.login().invoke(
+                new UserLogin("invalid-email@domain.com", "123456"))
+            );
+        } catch (RuntimeException e) {
+            throw e.getCause();
+        }
```
ignasi35: Two new tests.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -1,40 +1,91 @@
...
+        String password = "admin2";
+        userService = server.client(UserService.class);
+        UserRegistration userRegistration = new UserRegistration(name, email, password);
+        User createdUser = createNewUser(userRegistration);
+
+        Await.result(userService.login().invoke(new UserLogin(email, password)));
+        // no assertion required.
+    }
+
+    // TODO: replace with Forbidden.class
```
ignasi35: Should use `Forbidden` instead of `NotFound` (?)

ignasi35: (future PRs)

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
```diff
@@ -66,28 +64,26 @@ public UserServiceImpl(PersistentEntityRegistry registry, UserRepository userRep
...
+                                // TODO: replace with Forbidden.class
+                                throw new NotFound("Email or password does not match ");
                             }
-                                                  );
-           });
-
+                        } else {
+                            // TODO: replace with Forbidden.class
+                            throw new NotFound("User not found");
+                        }
```
ignasi35: Not only these two `NotFound` should be a `Forbidden`, we probably want them to use the same message (or none at all).

item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
```diff
@@ -41,7 +41,7 @@
  */
 public class ItemServiceImplIntegrationTest {
 
-    private final static Setup setup = defaultSetup().withCassandra()
+    private final static Setup setup = defaultSetup().withCassandra(true)
```
renatocaval: Why are we putting back the `withCassandra(true)`. In Lagom 1.4.0 we make this method less verbose by allowing `withCassandra()`. 

It doesn't make much sense to have `withCassandra(false)` because is the default, so `with` and `true` is redundant.

user-impl/src/test/java/com/example/auction/user/impl/UserServiceImplTest.java
```diff
@@ -1,40 +1,91 @@
...
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 public class UserServiceImplTest {
 
-    private final String name = "admin";
-    private final String email = "admin@gmail.com";
-    private final String password = "admin";
+
+    private final static ServiceTest.Setup setup = defaultSetup().withCassandra(true)
```
renatocaval: same as above

item-impl/src/test/java/com/example/auction/item/impl/ItemServiceImplIntegrationTest.java
ignasi35: You're right. I was getting compilation errors in IntelliJ and just went ahead. I had not reloaded the project in a while so it was was false negatives.

ignasi35:
> Didn't we say that we want to remove all this login stuff?

we did, that's why I timeboxed into at least having something that doesn't crash by default.

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
TimMoore: I always end up opening the `w=1` URL in a separate browser tab so I can flip back and forth for this reason.

user-impl/src/main/java/com/example/auction/user/impl/UserServiceImpl.java
TimMoore: It's a good thing we haven't implemented a way to update your email address! 😆

user-impl/src/main/java/com/example/auction/user/impl/UserRepository.java
ignasi35: Dear Github,

this is no working. Plesae fix ASAP as we needs for going to real productionz.

PR #170, opened 2018-03-16
Enable misc serializers for types known to have Java serialization defaults

ignasi35:
See https://github.com/lagom/lagom/issues/1239


TimMoore:
I wonder if we should turn off Java serialization entirely in the Online Auction examples. It would be an interesting experiment to try.

ignasi35:
> I wonder if we should turn off Java serialization entirely in the Online Auction examples. It would be an interesting experiment to try.

👍🏼

Adding new issue.

PR #183, opened 2018-04-11
Update link to LO4K8s

ignasi35:
Link redirection was already handled in techhub so users were already landing on the appropriate content, but as I was updating the text I thought I could update the link too.


TimMoore:
Thanks! The Scala version needs the same change: https://github.com/lagom/online-auction-scala/pull/77

PR #190, opened 2018-08-21
[WIP] gRPC integration

renatocaval:
This PR is in a very early stage. There is no real gRPC integration yet.

It depends on a current Lagom build of branch https://github.com/lagom/lagom/pull/1443.

My intention was primarily to test the ssl support. As we go, I (or someone else) can add a gRPC router to one of Lagom's service and consume it from another service.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ before_install:
   - sudo dpkg -i --force-confnew elasticsearch-5.0.2.deb
   - sudo service elasticsearch restart
 before_script:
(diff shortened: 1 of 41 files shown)


ignasi35:
@dwijnand I think we should first migrate the repository to lagom `1.5.0-M3` and on a separate PR (this?) integrate with `gRPC`.

dwijnand:
Sounds reasonable to me!

ignasi35:
This is abandoned

PR #200, opened 2018-11-23
Use Lagom 1.5.0 RC1

marcospereira:
Most of the changes are related to the webGateway. I also did some code cleanup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ before_install:
   - sudo dpkg -i --force-confnew elasticsearch-5.0.2.deb
   - sudo service elasticsearch restart
 before_script:
+  # As documented here: https://docs.travis-ci.com/user/database-setup/#elasticsearch
   - sleep 20
-  - sudo chmod +x /usr/local/bin/sbt # see https://github.com/travis-ci/travis-ci/issues/7703
 script:
   - sbt -DenableElasticsearch=true test
 cache:
--- a/build.sbt
+++ b/build.sbt
@@ -225,9 +225,20 @@ def elasticsearch: String = {
   }
 }
 
-def commonSettings: Seq[Setting[_]] = eclipseSettings ++ Seq(
+def evictionSettings: Seq[Setting[_]] = Seq(
+  // This avoids a lot of dependency resolution warnings to be showed.
+  // They are not required in Lagom since we have a more strict whitelist
+  // of which dependencies are allowed. So it should be safe to not have
+  // the build logs polluted with evictions warnings.
+  evictionWarningOptions in update := EvictionWarningOptions.default
+    .withWarnTransitiveEvictions(false)
+    .withWarnDirectEvictions(false)
+)
+
+def commonSettings: Seq[Setting[_]] = eclipseSettings ++ evictionSettings ++ Seq(
   javacOptions in Compile ++= Seq("-encoding", "UTF-8", "-source", "1.8"),
-  javacOptions in(Compile, compile) ++= Seq("-Xlint:unchecked", "-Xlint:deprecation", "-parameters")
+  javacOptions in(Compile, compile) ++= Seq("-Xlint:unchecked", "-Xlint:deprecation", "-parameters", "-Werror"),
+  scalacOptions += "-feature"
 )
 
 lagomCassandraCleanOnStart in ThisBuild := false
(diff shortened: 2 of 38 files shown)


.travis.yml
```diff
@@ -12,8 +12,8 @@ before_install:
   - sudo dpkg -i --force-confnew elasticsearch-5.0.2.deb
   - sudo service elasticsearch restart
 before_script:
+  # As documented here: https://docs.travis-ci.com/user/database-setup/#elasticsearch
   - sleep 20
-  - sudo chmod +x /usr/local/bin/sbt # see https://github.com/travis-ci/travis-ci/issues/7703
```
marcospereira: https://github.com/travis-ci/travis-ci/issues/7703 is now fixed.

build.sbt
```diff
@@ -225,9 +225,20 @@ def elasticsearch: String = {
...
+  // the build logs polluted with evictions warnings.
+  evictionWarningOptions in update := EvictionWarningOptions.default
+    .withWarnTransitiveEvictions(false)
+    .withWarnDirectEvictions(false)
+)
+
+def commonSettings: Seq[Setting[_]] = eclipseSettings ++ evictionSettings ++ Seq(
   javacOptions in Compile ++= Seq("-encoding", "UTF-8", "-source", "1.8"),
-  javacOptions in(Compile, compile) ++= Seq("-Xlint:unchecked", "-Xlint:deprecation", "-parameters")
+  javacOptions in(Compile, compile) ++= Seq("-Xlint:unchecked", "-Xlint:deprecation", "-parameters", "-Werror"),
```
marcospereira: `-Werror` to avoid deprecated APIs.

item-api/src/main/java/com/example/auction/item/api/DeliveryOption.java
```diff
@@ -1,16 +1,16 @@
 package com.example.auction.item.api;
 
+import lombok.EqualsAndHashCode;
 import lombok.Value;
 import java.util.Optional;
 import java.util.UUID;
 
-public abstract class DeliveryOption {
+public interface DeliveryOption {
```
marcospereira: Having these as abstract classes instead of interfaces was generating a Lombok warning:

```
Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
```

Also, in some places, we were using interfaces and at others abstract classes. This makes all of them consistent.

transaction-impl/src/main/java/Module.java
```diff
@@ -19,14 +19,5 @@
     protected void configure() {
         bindService(TransactionService.class, TransactionServiceImpl.class);
         bindClient(ItemService.class);
-
-        bind(TransactionServiceLifecycle.class).asEagerSingleton();
-    }
-
-    private static class TransactionServiceLifecycle {
```
marcospereira: This is not necessary anymore.

web-gateway/src/main/java/controllers/AbstractController.java
```diff
@@ -45,35 +39,34 @@ public AbstractController(MessagesApi messagesApi, UserService userService) {
         });
     }
 
-    private Nav getNav(PaginatedSequence<User> users, Optional<User> currentUser) {
-        return new Nav(messagesApi.preferred(Collections.emptyList()), users.getItems(), currentUser);
+    private Nav getNav(PaginatedSequence<User> users, Optional<User> currentUser, final Http.Request request) {
+        return new Nav(messagesApi.preferred(request), users.getItems(), currentUser);
```
marcospereira: get `Messages` from the `Request` instead of from an empty list.

web-gateway/src/main/twirl/views/createItem.scala.html
```diff
@@ -2,15 +2,15 @@
...
 
-@main(message("createItem")) {
+@main(messages("createItem")) {
 
     <h2>@message("createItem")</h2>
 
     @foundationForm(itemForm, routes.ItemController.createItem()) {
         <div class="column row">
-            @if(showInlineInstruction == true){
+            @if(showInlineInstruction){
```
marcospereira: 🙃

transaction-impl/src/main/java/Module.java
ignasi35: Is this the only `Module` where `gracefulShutdown` was still being used? I guess it was a leftover from the migration to 1.4.x...

ignasi35:
Done a smoke test locally (tested user handling, item creation, bidding, searching; not tested auction ending or transaction).

LGTM

PR #205, opened 2019-02-25
Bumps lagom and LBO

ignasi35:


ignasi35:
Not for merge. I just needed to trigger CLA validator.

PR #209, opened 2019-03-04
Lagom ElasticSearch Client

ihostage:
1. Use Lagom ElasticSearch Client
2. Fix pagination


ignasi35 (changes requested):
Hi @ihostage, I understand the principle of extracting code to make the purpose of each module more clear but I don't think that extracting to a separate library is a good idea.

Instead, I would have extracted the Elastic Search Client to a project in this same repo.

ihostage:
Hi, @ignasi35 😉 
Of course, we created a separate library for communicating with ES not for Online Action but for our many production services 😄
Just, we consider that Online Action is a good place to demonstrate this library so that members of the Lagom community do not waste time for copy&paste of code and can use the ready-made library.

octonato:
@ihostage, I understand your motivation, but on the other hand, if each author library comes here and modify the project to demo a library, the project loses its intention.

May I offer you a compromise? Instead, I suggest you build a Lagom app to demonstrate your library and we link it from the Lagom website. 

I will also suggest you to not build it based on online-auction, but instead have a dead simple app that concentrates on demonstrating how to use your ES client. Much like a recipe.

We have been discussing what we are going to do with the current recipes and samples we have. We need to reduce the number of side projects so we have more time to invest in Play and Lagom.

One thing is clear, we want to offer the community a place to expose their work and link integration libraries and I think your Elastic Search client is a good example of integration library that we should link.

PR #213, opened 2019-04-17
Bump to 1.5.0

ignasi35:
Supersedes #212

### DIFF
--- a/bidding-impl/src/main/resources/application.conf
+++ b/bidding-impl/src/main/resources/application.conf
@@ -21,3 +21,6 @@ akka.actor.serialization-bindings {
 }
 
 akka.actor.allow-java-serialization = off
+
+akka.discovery.method = akka-dns
+
--- a/build.sbt
+++ b/build.sbt
@@ -2,9 +2,6 @@ organization in ThisBuild := "com.example"
 
 scalaVersion in ThisBuild := "2.12.8"
 
-EclipseKeys.projectFlavor in Global := EclipseProjectFlavor.Java
-
-
 lazy val root = (project in file("."))
   .settings(name := "online-auction-java")
   .aggregate(
@@ -52,13 +49,14 @@ lazy val itemApi = (project in file("item-api"))
 
 lazy val itemImpl = (project in file("item-impl"))
   .settings(commonSettings)
-  .enablePlugins(LagomJava, SbtReactiveAppPlugin)
+  .enablePlugins(LagomJava)
   .settings(
     version := "1.0-SNAPSHOT",
     libraryDependencies ++= Seq(
       lagomJavadslPersistenceCassandra,
       lagomJavadslTestKit,
       lagomJavadslKafkaBroker,
+      akkaDiscoveryServiceLocator,
       cassandraExtras
     )
   )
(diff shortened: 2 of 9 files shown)


lightbend-cla-validator:
Hi @RustemZ,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

ignasi35:
Services using Akka's `ServiceDiscovery` don't declare a `method`. This needs fixing before merging.

ignasi35:
related https://github.com/lagom/online-auction-scala/pull/118#issuecomment-484231933

marcospereira (dismissed):
LGTM.

Should ` akka.discovery.method = akka-dns` be a default in `lagom-akka-discovery-service-locator`?

build.sbt
```diff
@@ -192,29 +194,26 @@ lazy val userImpl = (project in file("user-impl"))
...
-    // Uncomment the commented out line and remove the Scala line when issue #22 is fixed
-    EclipseKeys.projectFlavor in Global := EclipseProjectFlavor.ScalaIDE,
-    // EclipseKeys.createSrc := EclipseCreateSrc.ValueSet(EclipseCreateSrc.ManagedClasses, EclipseCreateSrc.ManagedResources)
-    EclipseKeys.preTasks := Seq(compile in Compile)
+    PlayKeys.playMonitoredFiles ++= (sourceDirectories in (Compile, TwirlKeys.compileTemplates)).value
   )
 
 val lombok = "org.projectlombok" % "lombok" % "1.18.4"
 val cassandraExtras = "com.datastax.cassandra" % "cassandra-driver-extras" % "3.6.0"
+val akkaDiscoveryServiceLocator = "com.lightbend.lagom" %% "lagom-scaladsl-akka-discovery-service-locator" % "1.0.0"
```
marcospereira, line 215: Hum, on a second pass: use javadsl instead.

marcospereira (commented):
Notice a small detail regarding the dependency.

bidding-impl/src/main/resources/application.conf
```diff
@@ -21,3 +21,6 @@ akka.actor.serialization-bindings {
 }
 
 akka.actor.allow-java-serialization = off
+
+akka.discovery.method = akka-dns
```
TimMoore, line 25: Doesn't it need other configuration for Cluster Bootstrap? Should production configuration be split into another file like `prod-application.conf`?

TimMoore (changes requested):
This seems to be incomplete. My guess is that deployment on Kubernetes hasn't been tested, because I don't think it will work as is. `README.md` and `KUBERNETES.md` still refer to Lightbend Orchestration.

If you want to split the work into separate pull requests to update for development and deployment, I think that's OK, but in that case, take out all of the deployment stuff entirely until that part is done. Otherwise it's actively misleading and people could waste a lot of time trying to figure out something that isn't expected to work.

TimMoore:
> Should `akka.discovery.method = akka-dns` be a default in `lagom-akka-discovery-service-locator`?

If it's going to set a default (I have requested that it should since the beginning) the best default to use would be [aggregate](https://doc.akka.io/docs/akka/current/discovery/#discovery-method-aggregate-multiple-discovery-methods) preconfigured with both [config](https://doc.akka.io/docs/akka/current/discovery/#discovery-method-configuration) and [DNS](https://doc.akka.io/docs/akka/current/discovery/#discovery-method-dns) (in that order of priority). That makes it really easy to plug in manual configuration to override DNS entries or add ones that aren't available in DNS.

ignasi35:
@TimMoore I've raised https://github.com/lagom/lagom-akka-discovery-service-locator/issues/81 to track that.

